PR #1, opened 2015-12-18
Parse xml string from RAM

Primordus:
Added support for parsing of an XML string from a Rust str (previously only bindings were provided to parse from a file). 
Also fixed the tests (hopefully in a correct way :smile:), cleaned up some warnings, probably a remainder from older Rust versions..

P.S. some tests seem to fail sometimes, may be worth looking into? (even if you hate writing them ;))

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,2 +1,5 @@
 language: rust
 rust: nightly
+addons:
+    apt:
+        packages: libxml2-dev
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,11 +2,40 @@
 name = "rustlibxml"
 version = "0.0.1"
 dependencies = [
- "libc 0.1.8 (registry+https://github.com/rust-lang/crates.io-index)",
+ "gcc 0.3.21 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.1.12 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "advapi32-sys"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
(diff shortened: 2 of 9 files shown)


Primordus:
I noticed the build on travis failed, the test cases seem to fail every now and then.. if you run the code yourself though and cargo test a few times, should work..
For example, I enabled travis on my fork and it succeeded there.. tests could probably use some serious refactoring..

dginev:
Hi @Primordus ! I didn't get a reply to my last email to @jfschaefer , so I am unsure if he has time to maintain this wrapper. For now I have gone and added the extra features I need to a fork here:
https://github.com/dginev/rust-libxml which is still incomplete but a few steps further along.

It ended up in a significant rewrite (you may be heading in a different direction). I'll add the equivalent changes to your PR changes there by hand, while we're waiting for @jfschaefer to take a look.

Primordus:
Hello @dginev, that's great news! 
I'll look into your fork of the code later, didn't pay much attention to what other people already did since I just needed this function for my application. 
This change might not be so difficult to port over because it's only 1 new rust binding for a libxml2 function :smile:!

dginev:
Hi again - indeed useful API methods to add!

I am quite happy to report I have integrated them with my fork, see commit: https://github.com/dginev/rust-libxml/commit/63f11995e83166d840b6b458f45bd6393c096cc1

I have refactored the wrapper to have the XML/HTML distinction happen as a property of the Parser object, which feels a bit cleaner. And made some of the setup/cleanup magic a bit more automatic, which seems to be largely working out as intended. Feel free to take a look and use that if you prefer.

Primordus:
Thanks, I will probably use your fork instead then! Hope @jfschaefer some day will pull in our changes :)..

dginev:
This PR got merged a bit indirectly, so I'll close here - the commits are already in master.

PR #3, opened 2016-02-23
Replace static mutex

ramn:
So that we can compile on stable rust

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@
 
 # Test results
 /tests/results/*.xml
+/tags
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: rust
-rust: nightly
+rust: stable
 addons:
   apt:
     packages: libxml2-dev
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 language: rust
-rust: stable
+rust:
+  - stable
+  - beta
+  - nightly
 addons:
   apt:
     packages: libxml2-dev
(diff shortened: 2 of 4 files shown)


ramn:
When I run the tests on my machine, they all succeed.

dginev:
Btw a good idea is to start using *both* stable and nightly on Travis from this PR on, and possibly even beta. After all the core idea is to make the crate work on more rust releases, rather than less.

dginev:
And the 3 projects I am involved with that use this crate are all based on the nightly toolchain, so breaking that here would be a no-go.

ramn:
What should I put in the travis file?

dginev:
Based on the [travis docs](https://docs.travis-ci.com/user/languages/rust) I'd suggest:
```
language: rust
rust:
  - stable
  - beta
  - nightly
```

ramn:
Thanks, I'll try that.

ramn:
Strange, works on 1.8 nightly on macos

dginev:
The joys of travis. I'll take a look a bit later today if you don't find an obvious cause for the travis failure.

ramn:
Can we set RUST_BACKTRACE=1 ?

dginev:
I just installed today's nightly and it broke not only your PR but also the main rust-libxml master. So not a problem for your PR, we should fix it in master first.

dginev:
FYI it seems that xmlReadFile and htmlReadFile are now returning null pointers in the test suite.

dginev:
@ramn I merged a fix that brings us up to date with the current nightly, would you care to rebase the PR on the latest master? I'll look into it then (Travis should pass)

ramn:
Rebased https://github.com/jfschaefer/rust-libxml/pull/5

Issue #2, opened 2016-02-23
Support Rust stable


ramn:
Would be nice to use this library from Rust stable. The StaticMutex is a problem. Do you think something like this could work instead? https://github.com/ramn/rust-libxml/tree/replace_static_mutex

I can make a pull request if you think the approach is ok.

dginev:
Hi @ramn , that sounds like a great idea.

If you already have the code written, feel free to file a PR for review. We should release on crates.io once the wrapper works with the stable Rust.

Btw @jfschaefer we should decide how to maintain the wrapper, and in which repo. We could also transfer it to the KWARC org, etc. The only thing to solve on that front is that the repo has an active maintainer at all times, so that it doesn't become stale.

ramn:
Here is a PR: https://github.com/jfschaefer/rust-libxml/pull/3

dginev:
Merged, thanks for contributing!

ramn:
Np, thanks for providing this great library!

PR #4, opened 2016-02-24
Compat for Rust nightly 22.02: hold CString::new results in var

dginev:
The basic issue here was that we were casting to a pointer as a chain call on ```CString::new```, which in the latest nightly seems to result into null pointers by the time they are passed into libxml.

The refacotoring here fixes that, and we should be able to merge the PR that adds support for Rust stable next ( #2 ).

@jfschaefer let me know if you don't want to maintain the repo / are short on time, if you give me commit access I can push directly into the repo. Or we can move it to the KWARC org, up to you.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,40 +2,17 @@
 name = "libxml"
 version = "0.0.2"
 dependencies = [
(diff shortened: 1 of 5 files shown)


dginev:
Oh, I actually already have commit rights here, merging.

PR #5, opened 2016-02-24
Replace static mutex

ramn:
Rebase of https://github.com/jfschaefer/rust-libxml/pull/3


dginev:
Looks good, thanks for refactoring away from the static mutex, much appreciated!

PR #8, opened 2016-06-28
Test, which shows bug in clone() (or drop?) function

grray:
Both tests gives me segmentation fault on linux with rust 1.9.0. I can't fix it myself.
Second test is more "real life", but I'm not sure, maybe compiler can optimize it and do it without cloning.

### DIFF
--- a/tests/base_tests.rs
+++ b/tests/base_tests.rs
@@ -173,6 +173,36 @@ fn test_class_names() {
 }
 
 #[test]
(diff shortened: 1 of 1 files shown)


dginev:
Thanks, will take a look! The global scope of memory management in libxml is admittedly a hard paradigm to wrap in Rust, not surprised that there can be a segfault. Great to have tests!

grray:
#10 fixes this

PR #9, opened 2016-06-28
Test, which shows xpath string() function bug

grray:
It can be done with get_property(), but it should work this way also, isn't it?

### DIFF
--- a/tests/base_tests.rs
+++ b/tests/base_tests.rs
@@ -173,6 +173,24 @@ fn test_class_names() {
 }
 
 #[test]
+/// Test that an xpath string() function processed correctly
+fn test_xpath_string_function() {
+  let parser = Parser::default_html();
+  let doc_result = parser.parse_file("tests/resources/file01.xml");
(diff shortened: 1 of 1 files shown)


dginev:
Yep, this also looks valid.

dginev:
Ok, I incorporated your test with a solution that requires an extra ```.to_string``` call on the result of evaluate to get the string. Getting 0 nodes was actually expected behavior, as the xpath returns a simple string. You can see the slightly modified use [here](https://github.com/KWARC/rust-libxml/commit/8aa50fc8325b0d0643c8e75f351e7ff487eb9706#diff-c113524fbe0a6383fda1359255d49c1bR184).

Down the line I expect to also add a ```.to_num``` call for getting out the results of XPath expressions meant to do counts and other numeric operations. But this should do for now - thanks for the test!

I'm off to release a 0.0.5 release, also with the updated license, to make all of your contributions publicly available. Feel free to send more PRs, much appreciated!

PR #10, opened 2016-06-29
Doc out of scope

grray:
First commit has test, which demonstrates segmentation fault, second commit fixes error and makes that test impossible to compile, so test is removed in second commit. This change also makes #8 impossible to compile, so that was the real cause of problem there.

### DIFF
--- a/src/xpath.rs
+++ b/src/xpath.rs
@@ -7,13 +7,15 @@ use std::ffi::{CString};
 
 ///The xpath context
 #[derive(Clone)]
-pub struct Context {
+pub struct Context<'a> {
     ///libxml's `ContextPtr`
     pub context_ptr : *mut c_void,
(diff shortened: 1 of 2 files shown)


grray:
Maybe even xpath::Context should own Document?

dginev:
Thanks! I am merging for now, just so that we have the latest fix in master.

Owning the entire document is a tricky choice, I would intuitively consider it wrong, as you may want to use many contexts on the same document (e.g. in subsequent different xpath calls), so consuming the document may be an issue. On the other hand since libxml is doing the actual memory management, our Rust clones of Document are cheap... But that makes things additionally confusing.

If the lifetime-based API to Context ends up too confusing, we can switch to ownership of Document with some extra documentation. I would need to do some more exhaustive tests before we figure that out, and possibly ship a minor release to cargo afterwards.

PR #11, opened 2016-06-30
Use the same options for parse_string() as in parse_file()

grray:
This is mainly for suppressing warnings, which libxml write to stderr, they really on the way in console application.


grray:
Oops, wrong commit

PR #16, opened 2017-11-02
Add ability to recursively remove ns

triptec:
Coming from ruby I've loved to use Nokogiri, so now that I'm porting some code I found that this wasn't in libxml but in Nokogiri but still think it is useful.

From the Nokogiri docs

> This could be useful for developers who either don't understand namespaces
or don't care about them.


dginev:
Sounds reasonable, thanks for the PR!

PR #17, opened 2017-11-07
Add import_node to Document

triptec:
I need to copy a node from one document to another and this seems to be the function I need.


dginev:
Looks great! Let me know if you also need crates.io minor releases, whenever tracking the master of the repository isn't sufficient anymore.

triptec:
I think I'll track the master for now.
Is it just me or is the xml manipulation libraries in rust a little immature?

dginev:
Well yes, indeed! Them being a little immature is the exact reason this wrapper exists - I am perfectly happy to see `rust-libxml` fade away into obscurity one day, but until the Rust libraries are as impressive as the old-but-gold libxml2 library, this wrapper can be a useful stopgap.

I am hoping the main Rust-native XML libraries are moving forward fast however, I haven't checked recently since my main XML munging needs are at present satisfied.

PR #18, opened 2017-11-09
Add attribute for gcc about unused variable

triptec:
As I'm using a local repo I get a warning each time I compile, this silence that warning a alas doesn't clobber my tiny console output.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -7,7 +7,7 @@
 extern crate gcc;
 
 fn main() {
-  gcc::Config::new()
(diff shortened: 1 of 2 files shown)


dginev:
I also updated the version of the gcc crate in Cargo.toml on master, since you seem to be using a newer capability. All good after :+1:

dginev:
d13bd0731ba53c3901cda416c99eefae5cf3cc0e

PR #23, opened 2017-11-24
Ergo

triptec:
Not suggesting to merge this but I've played around and I think I might be nice to be able to have ParseOptions and so on.

Mostly curious about what you think about the idea and where we would implement those api breaking things.


dginev:
Sorry for taking a while to follow-up here.

Can you explain a bit more what the "ergo" name stands to be, and the direction of the PR?

Moving to more Options structs could be nice, I'm quite fond of the technique in general and use them in other projects. Bindgen, as we discussed in the issue, is also nice to move towards.

triptec:
yeah, ergo was just a name I chose for the module as I wanted something more ergonomic ie. ergo::XML(File|Path|String), also I think there's a need for being able to set ParseOptions. I was also quite happy with parse_handler, together with parse_string and parse_file. Also there is a beginning of getting the errors from libxml2 into a Vec or errors but in the code only the message is used but it should use message errorcode, line number etc.

But when I started with DocumentFragment I didn't know what path I should take for the dom structure, I started looking at "arena" and how other crates implemented it and so on.

Now I think we should add some of those ergonomics, ParseOptions, better Errors to this crate but keep the "unsafe" dom structure. And in a new crate use this to make a safe rust idiomatic xml/html manipulation library.

dginev:
Agree, with the single exception of the word "safe" -- the current crate should be safe as a goal, even if it keeps its API a bit more basic and lacks ergonomics. So I'll try to work on that here.

PR #24, opened 2017-11-27
Free after unlink

triptec:
as per the docs http://xmlsoft.org/html/libxml-tree.html#xmlUnlinkNode, I found it with the leak sanitizer (https://github.com/japaric/rust-san)

### DIFF
--- a/src/c_signatures.rs
+++ b/src/c_signatures.rs
@@ -62,6 +62,7 @@ extern "C" {
   pub fn xmlNodeNs(cur: *const c_void) -> *mut c_void;
   pub fn xmlNextNsSibling(attr: *const c_void) -> *mut c_void;
   pub fn xmlUnlinkNode(cur: *const c_void);
+  pub fn xmlFreeNode(cur: *const c_void);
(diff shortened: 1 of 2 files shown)


dginev:
Hm, I think this may be a little trickier. If I remember correctly, unlinking detaches a node from its parent fragment, but keeps it in memory, as you may want to reattach it to a different fragment.

I have been fighting the global memory management in the past, and my memory is that the safest way to do things is to free the memory when the last Node pointing to a given pointer is `Drop`-ed in Rust. But unlinking shouldn't be an immediate free by default.

dginev:
I think the cause of the failing Travis test is precisely that tricky bit.

triptec:
Hmm, seems you are right, add a delete() or something? I do need to unlink and free =/

dginev:
You should make sure the Node drops correctly. The moment Rust has it out of scope the drop method should run and free the underlying libxml structure.

dginev:
I'll close the PR for now if you don't mind, just to avoid merging it broken by mistake. Feel free to reopen / make a new one.

Issue #25, opened 2017-11-28
Logic for drop


triptec:
So I think I rather like to know how we should solve this. Now there's no drop logic for nodes at all.

Lets start with my use case, I traverse a document, find node that I want to remove and call unlink() on them. As it is today those unlinked nodes gives a memory leak.

What I would propose is to add a pub fn remove(self){} to node, that would both unlink and free.

lweberk:
The drop could check for the existence of a parent and/or document reference pointer. 

If parent node is none (NULL), then we can safely assume we are a root node thus having to free ourself and our children (recursively done by xmlFreeNode). Otherwise the node is a child to another node in the tree which would be responsible of cleaning us and our children up instead.

What I could not figure out is, whether xmlUnlinkNode also severs the relationship of the node to the document context (`node->doc` relationship). It would be logical, but I'm not 100% on that. It is trivially testable though (by checking if `node->doc` pointer is NULL after unlinking).

If everything above holds, then `detach/unlink` with an implicit deletion via drop might be better? It gives both a way to "unlink and move/consume" as well as "unlink and ignore" with drop as end result, both in one interface.

triptec:
I spent some time reading the libxml2 source yesterday and what I found was that there wasn't a strait forward way to detect if a node has been detached/unlinked as there are nodes that haven't got a parent, namely namespace nodes. So that at least we would have to check. And maybe more.

I was kind of leaning to adding a detached or something to the Node struct but that doesn't seem right either so I just kinda gave up and have my own unlink that also frees. =(

dginev:
Did you already get a chance to see my attempt at implementing a drop for Documents in global.rs?
https://github.com/KWARC/rust-libxml/blob/master/src/global.rs

The idea there was that we can bookkeep on the Rust side of things when the last libxml object is dropped from Rust, so that we perform the global clean operations on the C-side.

I think that approach could be extended to bookkeeping individual pointers, or we can use something less homebacked such as [Rc](https://doc.rust-lang.org/std/rc/struct.Rc.html), to guarantee knowledge of the `final drop of a Node`.

[...]

PR #27, opened 2017-12-19
Minor updates to ownership and documentation

dginev:
This PR contains the usable upgrades from experimenting with ownership, related to #25 .

A much more ambitious refactor would be needed for a safe ownership model, as discussed in that issue, this minor release increases documentation and honesty of the current crate flaws.


src/tree.rs
```diff
@@ -705,10 +731,11 @@ impl Node {
   /// Creates a new `Node` as child to the self `Node`
   pub fn add_child(&mut self, child: Node) -> Result<Node, ()> {
     unsafe {
-      if xmlAddChild(self.node_ptr, child.node_ptr).is_null() {
+      let new_child_ptr = xmlAddChild(self.node_ptr, child.node_ptr);
+      if new_child_ptr.is_null() {
         Err(())
       } else {
-        Ok(child)
+        Ok(Node { node_ptr: new_child_ptr })
```
dginev: this was actually a bug, as there is subtle merging going on when adding e.g. several adjacent text nodes

PR #28, opened 2018-01-12
workaround free methods, v0.1.2

dginev:
Hey @triptec , let me know if you can quickly review this PR.

I very quickly added a couple of free methods for nodes and namespaces to allow for some workarounds until we can adapt your work for a 0.2 release and have the crate auto-free on drop.

Need to make some quick progress, as I recall you had to some time back, and this is the most conservative approach I could take - just adding a couple of methods

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,13 @@
 # Change Log
-## [0.1.2 (in active dev)]
+## [0.1.3 (in active dev)]
 
- * We welcome Andreas (@triptec) to the core developer team!
+## [0.1.2] 2018-12-01
+
+* We welcome Andreas (@triptec) to the core developer team!
+
+### Added
+ 
+* Workaround `.free` method for freeing nodes, until the `Rc<RefCell<_Node>>` free-on-drop solution by Andreas is introduced in 0.2
 
 ## [0.1.1] 2017-18-12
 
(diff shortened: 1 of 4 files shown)


dginev:
I'm also using VSCode's auto-formatting via rustfmt, which is the vebose part of the PR

triptec:
LGTM and rustfmt sounds like a good idea.

dginev:
Coo. I am debugging with valgrind on the repo experiencing the leaks, ensuring I get all the free methods I need in this PR and merging. Thanks for the quick review!

triptec:
I've never used valgrind but I used this when I was looking for leaks when I did my first naive implementation with free on unlink.
RUSTFLAGS="-Z sanitizer=$SAN" cargo run

though that doesn't work on OSX but it did work within a docker container and was quite simple.

dginev:
Cool, didn't know Rust had an in-built leak checker now! The valgrind call is pretty direct too, but I had to wrestle some allocator magic to make it work properly in my crate - so maybe I'll try the  rust flags way next time.

dginev:
Ok, my memory leaks are resolved with the changes here - and I snuck in a safer form of the casting to utf8 Rust strings for the tree module. Interestingly it was `get_content` that was leaking strings, rather than having entire Nodes going wrong.

Still, happy to have the upgrades.

PR #32, opened 2018-06-11
replace hardcoded include path with pkg-config

asmuth:


dginev:
Wonderful, thank you for contributing the PR!

PR #35, opened 2018-07-17
Enhance tree modularity

dginev:
For @triptec to quickly review - I went ahead and split out the four individual sub-modules of tree into their own Rust files, and organized them via `tree/mod.rs`.

The node implementation is still too large, but can last a bit longer as-is now. Feel free to take a quick look.

Aside: Additionally, I refactored 3 repositories that I have which work with rust-libxml to the latest master, and the edits were quite manageable - was done in 30 minutes with all 3. So the new API feels good-to-go in that regard. Still to profile for memory leaks there, but that will be my next PR.

### DIFF
--- a/src/tree/mod.rs
+++ b/src/tree/mod.rs
@@ -1,10 +1,10 @@
 //! The tree functionality
 //!
 
-pub mod document;
-pub mod namespace;
-pub mod node;
-pub mod nodetype;
+mod document;
+mod namespace;
+mod node;
+mod nodetype;
 
 pub use tree::document::Document;
 pub(crate) use tree::document::DocumentRef;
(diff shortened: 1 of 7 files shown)


src/tree/mod.rs
```diff
@@ -0,0 +1,13 @@
...
+pub mod document;
+pub mod namespace;
+pub mod node;
+pub mod nodetype;
+
+pub use tree::document::Document;
+pub(crate) use tree::document::DocumentRef;
+pub use tree::namespace::Namespace;
+pub use tree::node::Node;
+pub use tree::nodetype::NodeType;
```
dginev: I'm quite happy that this reexporting technique of `pub use` works so well in Rust. Because of it there is absolutely no breaking changes to the public-facing API, and the restructuring can be entirely under the hood.

dginev:
Thanks!

PR #37, opened 2018-07-19
Reuse bookkept nodes (towards enforced ownership)

dginev:
I am working on making the `ownership_flaws` test by @triptec behave as we expect it, with a strict ownership guard at least at runtime - via RefCell.

First step towards that is to make sure wrapping the same `node_ptr` returns the same reference-counted RefCell object.

While doing that I also noticed the `Drop` for nodes was done too high, and should instead be performed on the innermost `_Node` object, so that we only free unbound nodes once (a test caught that, which was awesome).


dginev:
The next step here - which I will give a try - seems to be submerging all mutable methods - in fact all libxml2-related methods - down to the `_Node` interface, and create aliases in the `Node` layer above, which performs a `self.0.borrow_mut()`.

Then we can guarantee that the `ownership_flaws` test will fail in runtime, which I think is "good enough", given the setup.

dginev:
Hm, except most of that was already achieved! I think the `Rc` hides the mutability conflict of two variables looking into the same RefCell. Will try to read up on common use patterns, see if there is a nice upgrade... 

I added a minor commit with extra vigilance for `.node_ptr_mut()` use, but it doesn't achieve anything particularly useful, aside from consistency.

dginev:
Ok, so the behavior in the ownership test is in fact a feature of the `Rc<RefCell<T>` approach, where it is intended by design to be able to mutate a value that is owned by multiple (and potentially immutable) owners:

https://doc.rust-lang.org/book/second-edition/ch15-05-interior-mutability.html#having-multiple-owners-of-mutable-data-by-combining-rct-and-refcellt

Still thinking on the exact design + tools we can use to improve ...

dginev:
@triptec I found a very good path to ensuring mutability guards of nodes that are being referenced multiple times (and are hence unsafe to mutate as per Rust's convention).

The technique requires improving one method in particular - `Node::node_ptr_mut` and you can see that here:
https://github.com/KWARC/rust-libxml/pull/37/commits/80e885b4cf12afbb15d7fdda107b5192edf2938b#diff-4be2264dc3ceae24480e433dc1ceafe1R93

As a result, a bit more of the API evolved, as the Result types became more important - a runtime mutability failure will be realized as an Error value returned by the method that encountered the conflict. 

This avoids built-in panics, so the libxml users can decide how to handle the errors (including to ignore them if so wished, which means the method calls become no-ops).

Here is your ownership test, now behaving as expected:
https://github.com/KWARC/rust-libxml/pull/37/commits/80e885b4cf12afbb15d7fdda107b5192edf2938b#diff-a85fb4258b82d429b1619791f0c144a4R31

I am quite happy with the solution, let me know what you think. Would be happy to release a 0.2.0 after this PR is merged, since I feel quite happy with the overall state of affairs.

triptec:
I think this is truly great, I was hoping we would get this but perhaps at a later time. Seeing this now makes me very happy. I’m sorry I haven’t got any time right now to test this out with my project, got a second daughter two days ago.

triptec:
Also I think we should release 0.2.0 and if we find more things to make this even better then let’s do that and release 0.3.0 etc

dginev:
Woohooo, congratulations!!! :tada: :balloon: :birthday: 

Let's ship `0.2.0` also in her honor then, I will merge and deploy to crates.io, port over my existing libraries and see if we get additional feedback.

Happy you are on board with the plan, now I leave you to celebrating and keeping up with the craziness!

PR #39, opened 2018-07-22
Full libxml2 coverage via bindgen

### LINKED ISSUE #19: Consider using bindgen
It might be useful to use [bindgen](https://rust-lang-nursery.github.io/rust-bindgen/) to generate an initial and comprehensive set of raw Rust bindings (in place of the incomplete hand-written bindings) and then refactor the lifecycle management wrappers to go around those.

dginev:
Fixes #19 

Interestingly, the swap was only slightly painful. Bindgen seems to have 3 obvious positives as an approach for us:
 * full coverage for free, without us having to maintain the `c_signatures` file.
 * transparent view into the C data structures, which gives us more power in the internal wrapper methods. This also means the low-level helper C methods we had can be redone in Rust.
 * the full type alias system of libxml2 is now available, which is a lot more readable (and sane) than casting to `c_void` pointers everywhere

There is more polish to be done here before the PR is merged, but it is already good for an early review and feedback


dginev:
Cool, this went well - this PR now removes all custom `.h` and `.c` code in the crate, and is pure Rust, from the bindgen functions and up. Feels a lot cleaner to not have any C directly deposited by the crate.

PR #43, opened 2018-09-26
Node::null, and clippy lints

dginev:
cc @triptec - another one to review.

I got lucky and found the trick to running valgrind in my other project (which keeps evolving as new features appear in nightly), more in [the issue](https://github.com/KWARC/llamapun/issues/20) for anyone interested.

And that revealed that using `Node::mock()` in an implementation of the `Default` trait is leaking memory - albeit tiny amounts.

So in this PR I finally did the C approach properly - I added `Node::null()` method, which uses a `ptr::null_mut`, and added an internal sibling of that method for `DocumentRef`. So now there is a memory-free placeholder node, which is more appropriate for my use case.

One could argue that the other project has an API mistake, and should really use an `Option<Node>` instead, but that makes the entire API way too cumbersome. In fact what I really want is a `PartialDefault` trait, as I would love to provide fallbacks for all fields except the node... But that's a brainstorm for another day.

I also added some recent clippy lint suggestions, and a new test. And naturally verified that no memory is leaked after refactoring to using `Node::null()`, which is happily indeed the case.

### DIFF
--- a/src/tree/document.rs
+++ b/src/tree/document.rs
@@ -81,6 +81,14 @@ impl Document {
     };
     Document(Rc::new(RefCell::new(doc)))
   }
+
+  pub(crate) fn null_ref() -> DocumentRef {
+    Rc::new(RefCell::new(_Document {
+      doc_ptr: ptr::null_mut(),
+      nodes: HashMap::new(),
+    }))
+  }
+
   /// Write document to `filename`
   pub fn save_file(&self, filename: &str) -> Result<c_int, ()> {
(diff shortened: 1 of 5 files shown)


triptec:
do you think you could add a few lines on how you are running valgrind to the readme or some other file if you have that fresh in mind. I've never used it and it might be good to know how when I work on this =)

dginev:
@triptec thanks for the suggestion, I added a new markdown file under tests/ for now, which can evolve further later. Merging and releasing a minor version next.

PR #44, opened 2018-12-06
update to 2018 edition of Rust

dginev:
The 2018 edition is now in Rust stable:
https://blog.rust-lang.org/2018/12/06/Rust-1.31-and-rust-2018.html

Decided to update my crates and version bump, adding this for a Travis check.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,6 @@ addons:
 after_success: ./scripts/doc-upload.sh
 env:
   global:
+    - RUST_TEST_THREADS=1
     - SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt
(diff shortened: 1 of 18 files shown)


dginev:
The update has revealed a test fragility (and potentially crate bug), with some intermittent segfaults when testing repeatedly locally. Managed to reproduce even with 1.31 stable:

```
test can_set_get_text_node_content ... error: process didn't exit successfully: `/home/deyan/git/rust-libxml/target/debug/deps/tree_tests-498d5e8b6ed99e44` (signal: 11, SIGSEGV: invalid memory reference)
```

Will try to find a patch before merging here.

dginev:
Good way to get the segfault reproducing on every test run is:
```
cargo test -- --test-threads 1
```

I have some ideas where things are going haywire by now, would need a bit more prodding to be certain...

dginev:
Issues I fixed: 
 * stopped deallocating the list returned by `.get_namespaces()` since I'm endlessly confused by the right timing to do so, may require additional bookkeeping on the Document level... postponing it for now
 * realized that `.import_node()` results in bookkeeping a Node twice - in both its prior document (from which it gets unlinked) and into its new receiver document. Which leads to two free() calls and a segfault. I added a `.forget_node()` helper to allow dropping a node when it gets imported somewhere new.

The `RUST_TEST_THREADS=1` is something I will keep enabled in Travis, as it helps expose fragility with memory usage.

Updating the documentation and merging here.

Issue #46, opened 2019-03-21
Basic thread safety


dginev:
Follow-up to #20 which didn't get too far. It would be great if we could have `Sync` nodes, so that we could e.g. do a rayon parallel iteration over a node's children, speeding up dramatically various read-only traversals.

What I don't currently understand is whether we must pay the price of an `Arc<Mutex<...>>` wrapper, or we could satisfy the guarantees in a lighter fashion. It would be great if I could find a different Rust wrapper that is already taking care of concurrency and learn from their implementation of `Send + Sync` guarantees for the wrapped pointers...

My current goal is to ensure thread-safety withing a *single* document, which should be substantially simpler than ensuring multi-document thread-safety. And just getting the read-only guarantees would still be a decent win, since that's most of what one needs in a huge class of applications (e.g. reading in config files, parsing statistics...)

dginev:
For posterity, writing down my conclusion of #47 and #53 

I have become convinced that parallel mutability over libxml2 is very likely more trouble than it is worth to retrofit. As the entire document tree is eagerly loaded in memory, to be truly safe in Rust one needs to recreate the entire tree as a set of lockable exclusion rules. Or to keep it simple - lock the entire document on each mutable operation.

As I have commented in #47, naively locking the document leads to long starvation blocking on any non-trivial number of threads, adding up to overhead that makes performance worse than the master branch single-threaded approach.

So, at least for the moment, I am happy to think of single threaded mutable `Node` and `Document`, which can't be used with parallel code, together with the parallelization-friendly read-only primitive `RoNode`, which allows for multi-threaded scanning.

Another note is that actual visible performance boost can be spotted only when processing huge batches of documents, or very large individual documents, so it wouldn't matter for more casual applications. 

On 32 logical threads (threadripper 1950x), using `RoNode` has shown 25-fold speedup on scanning a single 112 MB XML file, as well as a 20-fold speedup on scanning 1.2 million HTML files (see https://github.com/KWARC/llamapun/issues/28 ), down to only 2 hours from almost 2 days.

That said, since `0.2.10` the wrapper is now thread-friendly for all read-only use cases, so definitely *basic* thread safety is achieved.

PR #47, opened 2019-03-21
Thread-safe via `Arc<Mutex<...>>`

dginev:
First stab at thread safety ( #46 ).

I would be quite lucky if you have some time to review and think about the general problem @triptec . Or if you know of other crates that may have solved a similar issue nicely, that we can copy a method from.

In this PR I added the "best practice" from the official rust book, which is to hide the libxml node pointer behind even more indirection+guarantees:

```rust
type NodeRef = Arc<Mutex<_Node>>;
```

And you could still pretend this is a regular rust data structure and call `node.set_attribute(...)` or similar on a mutable node, without ever realizing there is an atomic reference counter and a mutex behind the `Node` struct. In theory this enables parallel traversals / mutations such as the mock test added here, to the tune of:
```rust
use rayon::prelude::*;
// ...
nodes.into_par_iter().for_each(|mut node| {
   node.set_attribute("key","val");
}
```

and get the parallel processing for free via the rayon iterator, while still remaining safe for mutation thanks to the `Arc<Mutex<...>>`.

The two main bits that are worrying me are:

 1. This loses performance for single-threaded applications, and I assume most crate users will focus on exactly that classic application type. Maybe it makes sense to have a separate set of parallel structs, such as `NodeAtomic` and `DocumentAtomic`, which implement Send+Sync, instead of overriding the basic structs, so that single-threaded users are not impacted... but the boilerplate will become huge if I try to maintain two implementations...

2. Sanity. I am quite troubled that I had to manually "claim" the final structs are Sync+Send via:
```rust
#[derive(Clone, Debug)]
pub struct Node(NodeRef);

unsafe impl Sync for Node {}
unsafe impl Send for Node {}
```

I would sleep much better if Rust could auto-derive those, based on the `Arc<Mutex<...>>` that is now wrapping `NodeRef`, but I am unsure if I am missing a piece to achieve that, or if that is not even possible for custom structs... Feedback welcome.

P.S. The rayon test, and rayon dependency, are just here for purposes of example, I would remove both before merging any version of this PR, as I don't think rayon should be a dependency for the generic libxml crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,15 +11,22 @@ license = "MIT"
 keywords = ["xml", "libxml","xpath", "parser", "parsing", "html"]
 build = "build.rs"
 exclude = [
-  "scripts/*"
+  "scripts/*","benches/*"
 ]
+
+[lib]
+name = "libxml"
 
 [dependencies]
 libc = "0.2"
-rayon = "1.0.0"
 
 [build-dependencies]
 pkg-config = "0.3.2"
 
-[lib]
-name = "libxml"
+[dev-dependencies]
+rayon = "1.0.0"
+criterion = "0.2.10"
+
+[[bench]]
+name = "parsing_benchmarks"
+harness = false
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ exclude = [
 
 [dependencies]
 libc = "0.2"
+rayon = "1.0.0"
 
 [build-dependencies]
 pkg-config = "0.3.2"
--- a/benches/parsing_benchmarks.rs
+++ b/benches/parsing_benchmarks.rs
@@ -3,8 +3,12 @@ extern crate criterion;
 
 use criterion::Criterion;
 use libxml::parser::Parser;
-use libxml::tree::Node;
+use libxml::tree::{Node, NodeType};
 use rayon::prelude::*;
+
+// -- workhorse functions
+// not *quite* classic depth-first search, since we keep all children at the current level in memory,
+// but certainly DFS-order for traversal
 
 fn dfs_single(node: Node) -> i32 {
   1 + node
@@ -22,10 +26,37 @@ fn dfs_parallel(node: Node) -> i32 {
     .sum::<i32>()
 }
 
+fn dfs_single_work2(node: Node) -> (i32, usize) {
+  if node.get_type() == Some(NodeType::TextNode) {
+    (1, node.get_content().len())
+  } else {
+    node
+      .get_child_nodes()
+      .into_iter()
+      .map(dfs_single_work2)
+      .fold((1, 0), |acc, x| (acc.0 + x.0, acc.1 + x.1))
+  }
+}
+
+fn dfs_parallel_work2(node: Node) -> (i32, usize) {
+  if node.get_type() == Some(NodeType::TextNode) {
+    (1, node.get_content().len())
+  } else {
+    let dfs_work = node
+      .get_child_nodes()
+      .into_par_iter()
+      .map(dfs_parallel_work2)
+      .reduce(|| (0, 0), |acc, x| (acc.0 + x.0, acc.1 + x.1));
+    (dfs_work.0 + 1, dfs_work.1)
+  }
+}
+
+// --- bencher functions
+
 fn bench_single_thread(c: &mut Criterion) {
   let parser = Parser::default();
   let doc = parser.parse_file("benches/big.xml").unwrap();
-  c.bench_function("single thread", move |b| {
+  c.bench_function("single thread DFS count", move |b| {
     b.iter(|| {
       let root = doc.get_root_element().unwrap();
       assert_eq!(dfs_single(root), 4_690_647)
@@ -33,10 +64,21 @@ fn bench_single_thread(c: &mut Criterion) {
   });
 }
 
+fn bench_single_thread_work2(c: &mut Criterion) {
+  let parser = Parser::default();
+  let doc = parser.parse_file("benches/big.xml").unwrap();
+  c.bench_function("single thread DFS count+length", move |b| {
+    b.iter(|| {
+      let root = doc.get_root_element().unwrap();
+      assert_eq!(dfs_single_work2(root), (4_690_647, 81_286_567))
+    })
+  });
+}
+
 fn bench_multi_thread(c: &mut Criterion) {
   let parser = Parser::default();
   let doc = parser.parse_file("benches/big.xml").unwrap();
-  c.bench_function("multi thread", move |b| {
+  c.bench_function("multi thread DFS count", move |b| {
     b.iter(|| {
       let root = doc.get_root_element().unwrap();
       assert_eq!(dfs_parallel(root), 4_690_647);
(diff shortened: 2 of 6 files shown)


dginev:
If I understand the Rust book correctly:
https://doc.rust-lang.org/nightly/book/ch19-01-unsafe-rust.html#implementing-an-unsafe-trait

If we have a raw pointer, irrespective of how many layers of guarantees we wrap it with, the compiler can never verify if it is Send+Sync, the way it could for native-derived types. Thus the onus is on the module author to manually make the claim via an `unsafe impl Send/Sync for X {}`, as I have done here.

In which case the main consideration is how to stay lean in performance?

dginev:
Got some encouragement that atomics aren't felt that strongly in single-threaded contexts, so maybe this is a sensible approach after all... Will run a benchmark on some large XML snippet to see if any visible slowdown occurs ...

dginev:
I had a sneaking suspicion this would be the case - a naive benchmark shows a parallel iterator DFS to be much slower than the single-threaded version. I pushed the example at b2fff0d2428c55b81ea53bddd6939894e45ef3f9

and the first benchmark results with Criterion were:
```
     Running target/release/deps/parsing_benchmarks-e718803c0af36e8a
single thread           time:   [1.4070 s 1.4151 s 1.4187 s]                           

multi thread            time:   [4.6913 s 4.6964 s 4.7012 s]                          
```

This is particularly shocking since I have 32 logical threads available on my 1950x threadripper CPU, so there may be a significant mistake in the way I've implemented the benchmark... Will dig in and report back. Also, this is not yet benchmarking the new "atomic" single-threaded performance vs the old `Rc` performance of master, gearing up to that next.

dginev:
Interestingly enough, I believe the slower performance comes from waiting for a Mutex lock to become available. The place where it seems most visible ("hottest path") is at `Node::wrap`, where we need a mutable lock over the document, to record the currently wrapped node pointer. Safety getting into the way of performance ...

The other aspect is that I am benchmarking a simple sum, which is a near-instant operation, and there is little load to split between the threads. If we make the assumption that instead each operation takes a more significant chunk of time, the multi-threaded code will appear a lot more useful.

triptec:
I might find some time this weekend

dginev:
Thanks! I'll do some more iterations on the PR until then. Out of curiosity I reran my rayon benchmark with 32,16,8,4,2 and 1 threads, and recorded the results in my last commit. It's a very explicit way to confirm that adding more threads adds more overhead, and my best estimate continues to be that the document `.lock` in `Node::wrap` is the main culprit for the waiting overhead.

One may wonder if we really need a mutex for threading within the same `Document` object, but omitting the mutex leads to an immediate runtime panick as the `RefCell` holding the document gets mutably borrowed multiple times, by each of the node threads.

So, we either need a redesign that makes `Node::wrap` lock-free (w.r.t document), or this PR will be useful only for jobs which have very significant thread-internal tasks, for which paying the overhead is still worth the thread speedup.

dginev:
Commenting out the two uses of `document` in `Node::wrap` (and thus losing memory safety!) I can confirm we're back to sane multi-threading territory:
```rust
single thread DFS count time:   [766.91 ms 768.93 ms 771.42 ms]                                  
                        change: [-46.972% -46.786% -46.604%] (p = 0.00 < 0.05)
                        Performance has improved.

single thread DFS count+length                                                                          
                        time:   [1.1354 s 1.1364 s 1.1374 s]
                        change: [-34.547% -34.357% -34.194%] (p = 0.00 < 0.05)
                        Performance has improved.

multi thread DFS count  time:   [692.04 ms 693.98 ms 695.69 ms]                                 
                        change: [-55.673% -55.559% -55.446%] (p = 0.00 < 0.05)
                        Performance has improved.

multi thread DFS count+length                                                                          
                        time:   [687.61 ms 688.83 ms 690.10 ms]
                        change: [-61.991% -61.896% -61.801%] (p = 0.00 < 0.05)
                        Performance has improved.
```

But even in the case of single-threaded classic use of the wrapper, with no rayon participation at all, once can spot a 30-40% speedup on this particular benchmark. Food for thought in the "price of safety" heading.

Cargo.toml
```diff
@@ -11,14 +11,22 @@ license = "MIT"
...
 libc = "0.2"
 
 [build-dependencies]
 pkg-config = "0.3.2"
 
-[lib]
-name = "libxml"
+[dev-dependencies]
+rayon = "1.0.0"
+criterion = "0.2.10"
```
dginev: the `[dev-dependencies]` header is pretty neat, it allows us to keep very heavy dependencies we'd like to experiment with for benchmarking, while completely ignoring them when `libxml` is used as an external dependency.

dginev:
Some thoughts while I continue to think on the data gathered here. The main reason for me starting this PR is that I have a collection of over 1 million HTML files that I need to very quickly iterate over and perform various tasks on. The task I am running this week is in fact very similar to the benchmark in this PR -- I do a traversal of the nodes, and record statistics on which node/attribute/value triples are in use, which then inform external tool development.

So my initial question was - can I speed a traversal up, knowing that I have all of the node data in memory (in C's libxml2), and knowing I have a read-only use case, where mutability wasn't even an issue. And it seems like the answer is nuanced, since we pay in runtime performance for certain aspects of safety... Then again, the bookkeeping design was designed in a single-threaded context, and may have a logical refactor that is both safe and fast in a multi-threaded use... Hm...

dginev:
So, I pushed a new branch with an entire new experiment that implements [my last comment](https://github.com/KWARC/rust-libxml/pull/47#issuecomment-475692805).

A good place to read into the branch is [here](https://github.com/KWARC/rust-libxml/compare/thread-safe-lock-on-mut?expand=1#diff-4be2264dc3ceae24480e433dc1ceafe1R130)

The two ideas being that 1) each mutable use of a Node pointer will enforce a document lock and 2) Nodes just wrap over their data, and the only `Arc<Node>` we need is in the Document bookkeeping, which is still vital for proper memory deallocation on Drop.

What I lost as a feature in that branch that if you get two rust `Node` objects that point to the same underlying C node, and call mutating methods on both of them, Rust will no longer realize this is a mutability conflict in runtime. This is the case since I had to go back (for now) to the simple wrappers we had before the current `Rc<Refcell<...>>`, in order to drop all the locking overhead and compare runtimes.

Ok, so benchmarking this lighter approach of bookkeeping/locking-on-mut, on the same examples as above, yields:
```
single thread DFS count time:   [459.57 ms 459.82 ms 460.33 ms]                                  

single thread DFS count+length                                                                          
                        time:   [710.99 ms 713.95 ms 717.42 ms]

multi thread DFS count  time:   [636.26 ms 637.30 ms 638.78 ms]                                 

multi thread DFS count+length                                                                          
                        time:   [619.51 ms 620.84 ms 622.04 ms]
```

Where we already see the multi-threaded version overtakes the single thread for the `count+length` payload, which is near-trivial, and indicates the gains will be significantly more visible in real-world applications. But more importantly, times are near 2x faster than the first naive implementation of the PR, which is quite considerable... 

I think I will sleep on the current state, and see if I can recover the `mutability_guards` test without big runtime penalties... Interesting!

triptec:
I haven’t had the time to look at those pr’s yet but I’ve been thinking about it a little and wonder a) do we optimize for reading or writing and considering that b) is there any crates that has other primitives than rwlock and mutex that might be a better fit for the problem. I’m thinking of parking_lot, crossbeam and others I haven’t heard of. I’ve never used any multithreaded things other than just using crates that might use it under the covers so yeah

dginev:
Certainly a deep and open-ended rabbit hole, since this is taking us beyond the simple wrap over libxml2, and we can take the crate in several different directions (and make several different kinds of mistakes...).

The ideal goal would be that you only pay for safety when you are doing something unsafe - so that clean reads, and orderly mutability work with no runtime cost associated, until you make a mistake -- or the next best formulation -- until you are at a mistake-prone piece of logic.

Any changes to the tree have to be guarded, but if I was designing a library from scratch, I would probably make it possible to add guards at different levels of granularity. Examples of what I would have in mind (in a perfect world):
 - If you are setting an attribute, only lock this node, allow all other mutability to take place in sibling threads
 - if you are CRUD-ing a child/subtree, lock only the affected subtree of the document, and allow sibling threads to have mutability anywhere else on the document fragment.
 - if you are reading, don't pay any runtime overhead for safety


Then a tree with N subtrees could have N threads do simultaneous localized changes at essentially constant time. And since I have already thought of that, any solution that I arrive for our current very humble wrapper crate feels like sour grapes - we end up paying various unnecessary overhead due to the "retroactive patching" design.

So back to humble goals, I would summarize as:
 - enable threading over libxml trees, e.g. ensure rayon iterators work reliably
 - enable "sane runtime overhead", where 1) single-threaded performance does not visible degrade from the current master and 2) it is *diffiuclt*, if not *impossible* to end up in a deadlock scenario, or a scenario where thread switching overhead visibly hurts performance.

I am starting to feel somewhat comfortable with the compromise solution on the new branch I made, but will still need at least a week of thinking and tinkering to feel "solid" about it ... 

As to using different synchronization primitives - maybe that helps towards getting closer to the "good design" goals? I will read a little into the alternative crates out there, I also haven't spent serious time studying the ecosystem.

dginev:
Closing this experiment in favor of #53 , will postpone thinking of parallel processing with mutability for a while longer (until someone comes and has a great solution!)

PR #48, opened 2019-03-24
Fix compiling on Mac OS X 10.14 (and presumably anywhere else that LIBXML_THREAD_ENABLED is set)

caldwell:
The problem is that `LIBXML_THREAD_ENABLED` changes the `xmlIndentTreeOutput` global from a simple global variable into a function call that returns an lvalue. The only proper way to detect and handle this is to just use the macro in C.

This patch adds c_wrappers.c and compiles it in the build.rs. I tested this with a recent nightly rust on Mac OS X 10.14 and Debian testing.

### DIFF
--- a/src/c_helpers.rs
+++ b/src/c_helpers.rs
@@ -95,7 +95,8 @@ pub fn xmlNodeGetName(cur: xmlNodePtr) -> *const c_char {
 
 pub fn setIndentTreeOutput(indent: c_int) {
   unsafe {
-    xmlIndentTreeOutput = indent;
+    let xml_indent_tree_output_ptr = __xmlIndentTreeOutput();
+    (*xml_indent_tree_output_ptr) = indent;
   }
 }
 


dginev:
Hi @caldwell , thanks a lot for this PR and problem report!

Would you mind helping me check if we can solve this using the `bindgen`-generated low-level helpers in `bindings.rs`, to avoid adding new C files?

Two questions:
 - Can you get the `.travis.yml` setup to fail in CI, by setting the env var, so that we can test this setup in the future? I can't seem to reproduce the issue locally for now on my Ubuntu setup.
 - Have you tried accessing the global state from `c_helpers.rs` via e.g.
```rust
pub fn setIndentTreeOutput(indent: c_int) {
  unsafe {
    let mut global_state = xmlGetGlobalState();
    (*global_state).xmlIndentTreeOutput = indent;
  }
}
```

I would consider it a win if we can keep the repo clean of `.c` files, we had a range of those early on, and bindgen was a huge help in clearing those out.

caldwell:
Hi, thanks for taking a look. I agree less C would be great. Your suggestion works on my setup, but I don't think it's quite right. In particular, after looking at the libxml2 sources, there are 2 places to get the globals. The struct returned from `xmlGetGlobalState()` is one of them, but there's also [a plain global int defined in `globals.c`](https://gitlab.gnome.org/GNOME/libxml2/blob/c494a0ba671188e5ab93593c568843087aa6583a/globals.c#L349). They have a function [`__xmlIndentTreeOutput()`](https://gitlab.gnome.org/GNOME/libxml2/blob/c494a0ba671188e5ab93593c568843087aa6583a/globals.c#L927) that chooses the struct or the global int based on whether it's running in the main thread ([which is always true if threads aren't enabled at all](https://gitlab.gnome.org/GNOME/libxml2/blob/c494a0ba671188e5ab93593c568843087aa6583a/globals.c#L31)). I was hesitant to use `__xmlIndentTreeOutput()` because it's named as an internal function and I wasn't sure if it was always there. But after looking at the code it seems to always be defined.

So I suggest this:

```rust
pub fn setIndentTreeOutput(indent: c_int) {
  unsafe {
    let xml_indent_tree_output_ptr = __xmlIndentTreeOutput();
    (*xml_indent_tree_output_ptr) = indent;
  }
}
```

This works on my Mac.

dginev:
Wonderful, thank you for taking a detailed look and refining the PR. Looks good to go! I'll release a minor version in case you want to have a cargo dependency.

PR #49, opened 2019-03-26
Fix c_helper.rs permissions

caldwell:
Apparently my Emacs saw the `#!` on the first line of the file and assumed it was a script, "helpfully" adding the `-x` permission. I didn't notice when I committed. Sorry about that.


dginev:
Subtle. I should redouble my efforts to review the PRs, didn't even think of monitoring permissions in my hurry.

PR #53, opened 2019-04-13
RoNode: read-only parallelization with minimal overhead

dginev:
A simple and pragmatic take on a multi-threaded primitive (see #47 ). 

A benchmark on a huge 112 MB XML file with this PR reveals:
```
single thread DFS count
  time:   [1.1032 s 1.1079 s 1.1120 s]                                     

single thread DFS count+length
  time:   [1.3465 s 1.3498 s 1.3532 s]

read-only single thread DFS count
  time:   [302.88 ms 303.53 ms 304.31 ms]

read-only single thread DFS count+length
  time:   [534.57 ms 535.03 ms 535.71 ms]

read-only multi thread DFS count
  time:   [33.617 ms 34.120 ms 34.625 ms]

read-only multi thread DFS count+length
  time:   [51.574 ms 52.269 ms 52.765 ms]

```

Where the single thread uses the Node implementation of master, while the `read-only` and multi thread benchmarks use the new `RoNode` construct.

The pragmatics of the PR were:
 * only solve parallel processing for one use case (= my current need), which keeps it simple and easy to understand
 * only pay for features you use mantra -- my use case is entirely one of "scanning" an existing document, so *no* mutation, *no* bookkeeping, *no* data races. 
 * As libxml eagerly loads the entire document in memory, it is safe to have concurrent read operations executing on the tree.
 * The PR is a conservative extension over master, which is unchanged, it only adds a new `RoNode` primitive and related logic for using it.
 * `RoNode` is just `Node` with all mutation and bookkeeping removed, including related methods. A bit of boilerplate overload to copy-paste it all, one could imagine it becoming a `NodeReader` trait instead.

As you can see from the sample benchmark, on a huge file with a very minimal compute job you get a 10x speedup when running using rayon parallel iterators on my 32 logical thread CPU.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,19 @@
 # Change Log
 
-## [0.2.10] (in active development)
+## [0.2.11] (in active development)
+
+## [0.2.10] 2019-14-04
+
+### Added
+
+ * `RoNode` primitive for simple and efficient **read-only** parallel processing
+ * Benchmarking a 120 MB XML document shows a twenty five fold speedup, when comparing `Node` to parallel rayon processing over `RoNode` with a 32 logical core desktop
+ * While `RoNode` is added as an experiment for high performance read-only scans, any mutability requires using `Node` and incurring a bookkeeping cost of safety at runtime.
+ * Introduced benchmarking via `criterion`, only installed during development.
+ * `benches/parsing_benchmarks` contains examples of parallel scanning via `rayon` iterators.
+ * added `Document::get_root_readonly` method for obtaining a `RoNode` root.
+ * added `Context::node_evaluate_readonly` method for searching over a `RoNode`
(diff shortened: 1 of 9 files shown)


dginev:
I would still want to add some tests here, and test on my actual application, before suggesting a merge, but welcoming feedback for this simplicity-seeking approach cc @triptec

dginev:
I have updated the benchmark and feel it's even more impressive to compare the `master` single-threaded performance to the read-only multi-threaded one -- 1.3s to 51ms, a twenty-five fold improvement in speed.

triptec:
Sounds good to me

dginev:
Thanks @triptec ! All existing code will remain unaffected, so feeling good about releasing the PR today, just pressing the buttons :+1:

dginev:
Shipped as 0.2.10 :tada:

Issue #33, opened 2018-06-17
XML with BOM not parsed


Gedweb:
I found trouble with some files, inside containing BOM.
It's just not parsed as a XML with `parse_file` method.

dginev:
Hi @Gedweb  - I am not personally familiar with BOM, and the wrapper currently only supports unicode documents - it's quite basic in that regard for now.

If you could point me to some libxml examples of working with BOM, that would help a lot. And you're even better off if you file a pull request with a fix that supports your workflow - we are quite good with merging enhancements in quickly and releasing new minor versions of the library.

Thanks for reporting the issue!

Gedweb:
The UTF-8 BOM just allows the reader to identify a file as being encoded in UTF-8.
After reading we can drop this bytes. I'll prepare pull request.

Gedweb:
So, inside libxml2 able `xmlSwitchEncoding` function, but I not so familiar in C.

dginev:
Hi @Gedweb ,

I took a second look - and I would need a minimal example of what you are trying to parse, and the Rust code you wrote to parse it and what you would expect to be different. As I mentioned earlier, just not that familiar with the BOM details.

The code you linked to in libxml should be executed exactly as you see it there, as the current wrapper doesn't set any encoding for the parser context internally.

Alexhuszagh:
I'm tagging myself to create an example shortly.

dginev:
I'm back to this issue, @Alexhuszagh , @Gedweb , just linking/attaching a standalone XML file and explaining the failure to parse will be a huge help for me to jump in and try to improve the wrapper.

Gedweb:
@dginev, look, first 3 bytes is the [BOM](https://en.wikipedia.org/wiki/Byte_order_mark).
BOM is not same for a different UTF encoding. When you found that magic octets in a start of file, just drop it. 
FYI, inside libxml2 you can found ready solutions. 
[cache.zip](https://github.com/KWARC/rust-libxml/files/3080919/cache.zip)
![image](https://user-images.githubusercontent.com/6544152/56145816-ad1f4780-5fad-11e9-86b2-4137c7c89eed.png)
In textual mode no one editor not show you that signs, only binary.

Alexhuszagh:
[...]

PR #55, opened 2019-06-14
Fix Issue #33

Alexhuszagh:
A comprehensive fix for issue #33. As I have not yet received any comments from this, I am tentatively submitting a PR for this patch. Please see the issue tracker for comments on this patch.


src/parser.rs
```diff
@@ -39,12 +43,84 @@ enum HtmlParserOption {
...
+  0
+}
+
+type XmlCloseCallback = unsafe extern "C" fn(
+  *mut ::std::ffi::c_void,
+) -> ::std::os::raw::c_int;
+
+
+///Convert usize to i32 safely.
+fn usize_to_i32(value: usize) -> Result<i32, XmlParseError> {
```
dginev: What is the motivation of this safe cast here? In particular I am wondering if we need to propagate it to other places in the wrapper, or it is contained to the parsing logic.

dginev: If it's indicating known cases where libxml wouldn't be able to handle an input of too large size, it may be better behavior to, for example, return an error indicating the overflow, rather than do a safe trim.

dginev: Cool handler, just reread and saw the Result type. I wonder if a more explicit name would be `try_usize_to_i32` to indicate the `Result` return type, but that's just thinking out loud for now.

dginev:
I'm very grateful for the extended test suite, thanks for the upgrade!

src/parser.rs
Alexhuszagh: I like that a lot, actually. It fits with the naming conventions from the TryFrom trait, making it signify its intent exactly.

PR #56, opened 2019-06-17
[minor] cleanup of new Parser code

dginev:
@Alexhuszagh here is an incredibly minor patch for what I was referring to as cleanup. Just the bits that struck me immediately, I'm happy to postpone any further refactoring until a need presents itself.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -5,10 +5,12 @@ use crate::c_helpers::*;
 use crate::tree::*;
 
 use std::convert::AsRef;
-use std::io;
+use std::ffi::c_void;
 use std::ffi::{CStr, CString};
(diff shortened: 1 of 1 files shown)


Alexhuszagh:
> @Alexhuszagh here is an incredibly minor patch for what I was referring to as cleanup. Just the bits that struck me immediately, I'm happy to postpone any further refactoring until a need presents itself.

That looks great, I for some reason thought you used a style in C++ to avoid ADL, which would explain the unambiguous declarations of `c_void` and `c_int`. Looking back, I can't see where I got that erroneous idea, so this cleans up the code a lot.

dginev:
Cool, thanks for clarifying!

PR #59, opened 2019-09-26
Document to string with save options

### LINKED ISSUE #58: Document to string with xmlSaveOptions
I would very much welcome some way to serialize a Document that is able to take `xmlSaveOptions` into account.
I hacked together a branch of rust-libxml locally that does exactly what I need for my application:

```
  pub fn to_string_custom(&self) -> String {
    unsafe {
      // allocate a buffer to dump into
      let buf = xmlBufferCreate();
      let c_utf8 = CString::new("UTF-8").unwrap();
      let options = xmlSaveOption_XML_SAVE_NO_EMPTY + xmlSaveOption_XML_SAVE_AS_HTML;

      let save_ctx = xmlSaveToBuffer(buf, c_utf8.as_ptr(), options as i32);
      let _size = xmlSaveDoc(save_ctx, self.doc_ptr());

      let result = xmlBufferContent(buf);
      let c_string = CStr::from_ptr(result as *const c_char);
      let node_string = c_string.to_string_lossy().into_owned();
      xmlBufferFree(buf);

      node_string
    }
  }
```

It would be cool to somehow generalize this and upstream it. Since there is already `to_string()` that uses `xmlDocDumpFormatMemoryEnc` I'm not sure how something that uses the `xmlSaveCtxt` would fit into the picture.

Is the functionality something that could find its way into the code base? And if so, how would you implement it?
I'd like to submit a PR, but think I need some guidance on how this should be implemented.

jangernert:
- fixes #58 
- Add a second method to serialize a document with `xmlSaveOption`: `to_string_with_options`
- remove format option from `to_string` since it is covered by the new method
- small whitespace and ordering changes were the result of a `cargo fmt` and can be reverted if needed


src/tree/document.rs
```diff
@@ -171,26 +213,51 @@ impl Document {
   }
 
   /// Serializes the `Document`
-  pub fn to_string(&self, format: bool) -> String {
+  pub fn to_string(&self) -> String {
```
dginev: You can entirely remove the `to_string` definition here, it is auto-generated by the `fmt::Display` implementation

src/c_helpers.rs
```diff
@@ -93,13 +93,6 @@ pub fn xmlNodeGetName(cur: xmlNodePtr) -> *const c_char {
   unsafe { (*cur).name as *const c_char }
 }
 
-pub fn setIndentTreeOutput(indent: c_int) {
-  unsafe {
-    let xml_indent_tree_output_ptr = __xmlIndentTreeOutput();
-    (*xml_indent_tree_output_ptr) = indent;
-  }
-}
-
```
dginev: Excellent to see this removed, thanks!

dginev:
Thanks @jangernert ! This looks good to merge after you simply delete the explicit `to_string` implementation and let `fmt::Display` add it in for us.

src/tree/document.rs
jangernert: Done

jangernert:
Cool! Happy to help.

PR #62, opened 2019-12-05
Implements std Error trait impl for XmlParseError

lweberk:
For error chaining in depending libraries or applications.


dginev:
Thank you for that upgrade!

Issue #72, opened 2020-04-19
Fails to build on windows 


coding-yogi:
I am using libxml as a dependency for one of my projects but it fails to build on windows. Any pointers to resolve this issue?

```
error: failed to run custom build command for `libxml v0.2.14`

Caused by:
  process didn't exit successfully: `C:\Users\xyz\work\rust\bombardier\target\release\build\libxml-77d8a84e5c34f954\build-script-build` (exit code: 101)
--- stderr
thread 'main' panicked at 'Could not find libxml2 using pkg-config', C:\Users\xyz\.cargo\registry\src\github.com-1ecc6299db9ec823\libxml-0.2.14\build.rs:5:5
```

dginev:
Short answer is that the crate has never been used under windows, so it is expected behavior for it to fail on a first try. I haven't actually even used rust itself under windows, so I wouldn't be able to offer any patches on a short notice. Happy to merge in any PRs that offer upgrades in this direction however, always welcome!

coding-yogi:
Thanks, I have no clue why it fails so unfortunately cannot submit any PR. This is the only lib which I found to be working well in case of XML and HTML xpath parsing.

Only thing I tried is downloading libxml2 for windows and adding it to path with no luck

coding-yogi:
Somehow got it built on windows using MSYS2 and mingw64 , is it possible to statically link libxml2 so that on windows I do not need the libxml2 and its dependecies?

dginev:
Good to hear @coding-yogi . And I wouldn't bundle in libxlm2 directly no, definitely not a fan of carrying it around. I am instead waiting on the complete rust transpilation of libxml2 by [c2rust](https://c2rust.com/manual/examples/libxml2/index.html). We've had some discussions with them that they may either publish that as a standalone crate or bundle it with this wrapper, when they feel ready, so that may be a roundabout way of getting simpler windows builds.

coding-yogi:
Good to know, will close this issue for now

simoin:
I've tested rust-libxml built on windows with `vcpkg` crate. [https://github.com/simoin/rust-libxml/blob/win/build.rs](url)
just add this to build.rs, or make `pkg-config`/`vcpkg` into feature:

```rust
if pkg_config::find_library("libxml-2.0").is_ok() {

} else if vcpkg::find_package("libxml2").is_ok() {

} else {
    panic!("Could not find libxml2")
}
```
![](https://i.imgur.com/zOYPTC1.png)

[...]

PR #77, opened 2020-09-28
#72-add-vcpkg-optional-dependency

JoshuaNitschke:
add opt in dependency for vcpkg
make pkg-config optional but default

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,8 +20,12 @@ name = "libxml"
 [dependencies]
 libc = "0.2"
 
+[features]
+default = ["pkg-config"]
+
 [build-dependencies]
-pkg-config = "0.3.2"
+pkg-config = {version = "0.3.2", optional = true }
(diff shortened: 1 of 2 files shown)


dginev:
Thanks, this looks great! Is there some smart trick to make vcpkg default if the current OS is windows? Would be great to have the windows builds succeed by default. Happy to merge here in either case, definitely a great upgrade.

JoshuaNitschke:
@dginev - updated.  I clones this commit, ran `cargo test`, and it built on:
- Windows 10
- WSL (Debian)
- OSX 10.15.4

JoshuaNitschke:
Might be failing the windows checks because vcpkg isn't installed?

dginev:
Perfect, thanks @JoshuaNitschke !

I just updated appveyor, and will fix the builds in master, your code already looks spot on, thanks for setting everything up. Once I get appveyor happy I'll ship a minor release of the crate.

PR #84, opened 2021-05-13
Allows to find libxml2 based on env variable if pkg config is not present

przygienda:


dginev:
Why wouldn't it be available? It's currently a hard requirement in [Cargo.toml](https://github.com/KWARC/rust-libxml/blob/master/Cargo.toml#L30), or respectively `vcpkg` under windows. I think to have a working Cargo build with only an env variable, we'd need a separate feature guard so that you can still build using cargo.

przygienda:
it builds using Cargo that way but I'm here on a system that does not have
pkg config in place (good amount of embedded systems in fact) so setting
the env variable is the only way.

dginev:
Got it, that makes sense! I haven't worked on such systems myself, so the explanation was necessary - thanks! I'll merge here and update the various doc files and push a minor release out.

przygienda:
appreciated, thanks. otherwise the crate works perfectly. thanks for the
work. tony

przygienda:
BTW, 0.2.17 broke 0.2.16 APIs, you may want to move to 0.3, especially the
parse_string variants

dginev:
A bit of an unsatisfying 0.3.0 release, but for now I'll just run with your suggestion to make it a bit more obvious that there may be refactoring needed. Thanks!

I think in the future I'll add deprecation warnings when we refactor single methods, via the `#[deprecated(...)]` attribute, and avoid the need for instant version bumps up.

przygienda:
the deprecation stuff is nice but not necessary really. keeping the 0.x+1
and then 1.x+1 notation to indicate that APIs changed is good enough.

thanks for publishing. 0.3.x works like a charm on 3 different platforms
including the one that doesn't have pkg-config on it.

PR #86, opened 2021-05-21
Add ParserOptions

### LINKED ISSUE #85: Allow user-specified ParserOptions
My use-case is parsing an HTML fragment, manipulating it and then serializing it again as a fragment. Currently there is no way around serializing it as a full HTML document and having to remove the "chrome" again afterwards:
```
<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN" "http://www.w3.org/TR/REC-html40/loose.dtd">
<html><body>
</body></html>
```

As far as I understand `HTML_PARSE_NOIMPLIED = 8192, // Do not add implied html/body... elements` would solve this for me. But the crate doesn't offer the option to pass options for parsing. There is however a comment in the code suggesting that the only reason that this didn't happen is time/manpower.
https://github.com/KWARC/rust-libxml/blob/master/src/parser.rs#L201

If it is okay with you @dginev I would like to implement this in a similar fashion as #59.

edit: there is also the workaround of creating a new empty document and moving all the relevant nodes from the old to the new doc. But it's not super obvious and a bit cumbersome.

jangernert:
Fixes #85

To avoid too many `parse_file_with_` variants I included the encoding into the `ParserOptions`. Hope that is acceptable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [0.2.18] (in development)
 
+* Change `Parser::parse_file/string_with_encoding` to `Parser::parse_file/string_with_options`. Introduce `ParserOptions` which encapsulates the forced encoding setting together with libxml2s HTML and XML parser options.
+
 ## [0.2.17] 2021-17-05
 
 ### Added
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -77,16 +77,14 @@ impl<'a> ParserOptions<'a> {
       (
         $condition:expr => $variant:ident
       ) => {
-          {
-            if $condition {
-              match format {
-                ParseFormat::HTML => HtmlParserOption::$variant as i32,
-                ParseFormat::XML => XmlParserOption::$variant as i32,
-              }
-            } else {
-              0
-            }
+        if $condition {
+          match format {
+            ParseFormat::HTML => HtmlParserOption::$variant as i32,
+            ParseFormat::XML => XmlParserOption::$variant as i32,
           }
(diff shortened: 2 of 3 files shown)


dginev:
Looks good actually! The only comment is that since `ParserOptions` only contains booleans and a borrowed str, it may be more idiomatic to pass it in as owned data, rather than as `&ParserOptions`. From what I recall Rust prefers cloning booleans to passing them by reference (it's faster), so cloning a `ParserOptions` struct ought to be fine, whenever needed.


 Wondering if you'd like to add a few lines explaining the changes to [CHANGELOG.md](https://github.com/KWARC/rust-libxml/blob/master/CHANGELOG.md) as well? I can do after merging if you prefer.

jangernert:
Passing a reference to `ParserOptions` is just out of habit. But you're right: in this case there no benefit. What do you prefer?

I can write a changelog entry. Should I describe all the options or simply mention the possibility to pass options and link the the libxml2 docs?

dginev:
Yep, I'd prefer passing the owned struct, thanks!

As to the docs - no need for too much detail, but mentioning the public Rust method+struct would be good (mentioning adding `ParserOptions` and `parse_file_with_options`, `parse_string_with_options`, while removing `parse_file_with_encoding`, `parse_string_with_encoding`), especially since existing code may need to be refactored.

jangernert:
Done!

dginev:
Wonderful, thanks for the upgrades! Much appreciated.

PR #87, opened 2021-06-18
fix typos in docs

TianyiShi2001:
Thank you for creating and continuing to maintain this brilliant crate! It is the only solution that allows me to seemlessly translate my `lxml`-dependent Python code, with a significant usage of XPath (on HTML), into Rust (I have no idea why it is less popular than the other buggy and functionally incomplete crates; although porting from C is less exciting than writing in pure Rust, I think being able to work robustly is more important).

I'm currently not familiar with C (and its porting to Rust) , but I'm learning. I look forward to contributing to the source code when I become competent in the future. For now, what I can do is to fix some typos in the docs and provide any bug reports.

### DIFF
--- a/src/readonly/tree.rs
+++ b/src/readonly/tree.rs
@@ -178,7 +178,7 @@ impl RoNode {
     NodeType::from_int(xmlGetNodeType(self.0))
   }
 
-  /// Returns true iff it is a text node
+  /// Returns true if it is a text node
   pub fn is_text_node(self) -> bool {
     self.get_type() == Some(NodeType::TextNode)
   }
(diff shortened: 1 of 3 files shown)


dginev:
Thanks for the housekeeping PR @TianyiShi2001 , and for the kind words!

I am certainly looking forward to having a libxml-parity library in pure idiomatic Rust, as there is a lot that could be done better if one starts from scratch. 

But, similarly to what you describe, I needed the production-ready state of libxml as a foundation, so this wrapper was born. We'll keep it around for as long as we need it.

One exciting avenue we were planning on trying is actually making the crate self-contained and use a Rust transpilation of libxml by [c2rust](https://c2rust.com/manual/examples/index.html). Unclear if that will work out, but it's a curious idea.

PR #89, opened 2021-08-02
Add MacOS and Windows installation steps to README

dginev:
Hi @hurlenko and @lepapareil ,

Could you take a brief look if adding the installation dependencies to the README in this PR looks reasonable? I've kept references in so that people can track the original discussions as well, when necessary.

Thanks again for contributing these, and I hope folks find them helpful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [0.3.1] (in development)
 
+* Added: Thanks to @lepapareil, @hurlenko and @ballsteve for contributing installation docs for Windows and MacOS.
+
 ## [0.3.0] 2021-27-05
 
 * Change `Parser::parse_file/string_with_encoding` to `Parser::parse_file/string_with_options`. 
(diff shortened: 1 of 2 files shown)


lepapareil:
Ok for windows 👌

hurlenko:
Looks good to me 🙂. Also the chocolatey steps seem to be optional, and are only meant to be used to install rust. At least I skipped them and compiled libxml on Windows successfully:

```
C:\> choco install -y curl git unxUtils winlibs-llvm-free
C:\> refreshenv
C:\> %ChocolateyInstall%\bin\curl --output "c:\rustup-init.exe" "https://static.rust-lang.org/rustup/dist/i686-pc-windows-gnu/rustup-init.exe"
C:\> c:\rustup-init.exe -y --default-toolchain stable-x86_64-pc-windows-msvc
```

Thanks!

dginev:
Good point, I can trim it down a bit further. The new instructions will go to crates.io with the nest release, thanks both for the quick reviews 👍🏻

PR #90, opened 2021-08-02
Remove tests using null pointer deref

dginev:
The recent versions of rustc are throwing warnings at some bindgen-generated tests, such as:

```
warning: dereferencing a null pointer
     --> src/bindings.rs:17692:15
      |
17692 |     unsafe { &(*(::std::ptr::null::<_xmlSchemaNotation>())).type_ as *const _ as usize },
      |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this code causes undefined behavior when executed
      |
      = note: `#[warn(deref_nullptr)]` on by default
```

This PR simply comments all of those tests out, as an experimental way out. I prefer that to disabling the warning, since the technique sounds dangerous enough to not rely on in general.

### DIFF
--- a/src/bindings.rs
+++ b/src/bindings.rs
@@ -614,16 +614,16 @@ fn bindgen_test_layout___fsid_t() {
     4usize,
     concat!("Alignment of ", stringify!(__fsid_t))
   );
-  assert_eq!(
-    unsafe { &(*(::std::ptr::null::<__fsid_t>())).__val as *const _ as usize },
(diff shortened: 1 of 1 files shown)


dginev:
I'll merge these for now and hope not having them doesn't hide some obscure problems on less used platforms. The (several screens of) undefined behavior warnings are a bit too much to keep in master.

PR #93, opened 2022-01-05
Wrapper for xmlXPathCompile binding

bcpeinhardt:
Haven't written any proper tests to see what the function covers, but getting back an error msg of "invalid xpath syntax" is promising!

### DIFF
--- a/src/xpath.rs
+++ b/src/xpath.rs
@@ -259,6 +259,8 @@ pub fn xml_xpath_compiles(xpath: &str) -> bool {
   let c_xpath = CString::new(xpath).unwrap();
     let xml_xpath_comp_expr_ptr = unsafe { xmlXPathCompile(c_xpath.as_bytes().as_ptr()) };
     let result = !xml_xpath_comp_expr_ptr.is_null();
-    unsafe { libc::free(xml_xpath_comp_expr_ptr as *mut c_void); }
+    if result {
+      unsafe { libc::free(xml_xpath_comp_expr_ptr as *mut c_void); }
+    }
     result
 }
--- a/src/xpath.rs
+++ b/src/xpath.rs
@@ -251,3 +251,14 @@ impl fmt::Display for Object {
     }
   }
 }
+
+/// Calls the binding to http://xmlsoft.org/html/libxml-xpath.html#xmlXPathCompile and return true if 
+/// a non-null pointer is returned. The idea is to use this to validate an xpath independent of context.
+/// Tests describing what this validates in tests/xpath_tests.rs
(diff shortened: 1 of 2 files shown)


bcpeinhardt:
^^ Regarding my earlier commit message stack overflow tells me `free` on a null pointer is a no-op not a double free but it sure _feels_ wrong haha.

dginev:
Thanks for contributing this @bcpeinhardt ! I was extra-slow in getting to replying to this PR, apologies.

One bit we can address here is the name of the new method. For example, we have a parser method called `is_well_formed_html`, which is closer to what I would expect of a condition-checking function returning a boolean. 

Maybe `is_well_formed_xpath` would be a name that is easier to read in vacuum?

src/xpath.rs
```diff
@@ -251,3 +251,16 @@ impl fmt::Display for Object {
...
+/// a non-null pointer is returned. The idea is to use this to validate an xpath independent of context.
+/// Tests describing what this validates in tests/xpath_tests.rs
+pub fn xml_xpath_compiles(xpath: &str) -> bool {
+  let c_xpath = CString::new(xpath).unwrap();
+    let xml_xpath_comp_expr_ptr = unsafe { xmlXPathCompile(c_xpath.as_bytes().as_ptr()) };
+    let result = !xml_xpath_comp_expr_ptr.is_null();
+    if result {
+      unsafe { libc::free(xml_xpath_comp_expr_ptr as *mut c_void); }
+    }
+    result
```
dginev, line 265: I think a slightly more idiomatic rust way of stating the ending here would be:

```rust
if xml_xpath_comp_expr_ptr.is_null() {
  false
} else {
  unsafe { libc::free(xml_xpath_comp_expr_ptr as *mut c_void); }
  true
}
```

bcpeinhardt: I agree this is "rustier" and generally easier to read! As far as the function name, I chose it to match the c function. You are correct that the semantic reason I have for "compiling" the xpath is to check that it is well formed, but I didn't want to make that assertion without confirming it. If you can confirm that a non null return from `unsafe { xmlXPathCompile(c_xpath.as_bytes().as_ptr()) };` guarantees a well formed xpath, then I also agree the function name should be changed.

dginev: I will change the function name in a separate PR and merge there, since I would like to include the functionality, but we don't have to agree on my naming reasoning.

The way I see it, it is not for me to *guarantee* the XPath is well-formed, but for cases where it isn't - but does compile - for users to report bugs to the libxml2 team. The intention of a compiled XPath is that the libxml2 engine has figured out some utility for employing it.

PR #94, opened 2022-01-19
Box errors with Send and Sync

jameswald:
Allows errors to be sent between threads safely. Fixes errors such as:

```
`(dyn std::error::Error + 'static)` cannot be sent between threads safely
    |
    = help: the trait `std::marker::Send` is not implemented for `(dyn std::error::Error + 'static)`
    = note: required because of the requirements on the impl of `std::marker::Send` for `std::ptr::Unique<(dyn std::error::Error + 'static)>`
    = note: required because it appears within the type `std::boxed::Box<(dyn std::error::Error + 'static)>`
```

### DIFF
--- a/src/tree/namespace.rs
+++ b/src/tree/namespace.rs
@@ -18,7 +18,7 @@ pub struct Namespace {
 
 impl Namespace {
   /// Creates a new namespace
-  pub fn new(prefix: &str, href: &str, node: &mut Node) -> Result<Self, Box<dyn Error>> {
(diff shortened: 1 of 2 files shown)


dginev:
This upgrade seems extremely reasonable, thank you for contributing it @jameswald !

I'm now realizing I haven't yet moved this repository over to github actions, so I get no CI confirmation on PRs, which is what I'll run off and do just after I merge here.

Feel free to send anything else that seems useful on your end over 🚀

jameswald:
Thanks for the rapid merge, greatly appreciated!

PR #97, opened 2022-03-26
Add support for has_attribute, has_attribute_ns

dginev:
Another cosmetic upgrade, allowing to use the just slightly faster [xmlHasProp](https://web.mit.edu/afs.new/sipb/user/yoz/libxml2-2.7.3/doc/html/libxml-tree.html#xmlHasProp) libxml2 function when only interested in whether an attribute exists or not - without allocating the value in Rust.


dginev:
Merging here to get a release out later today.

PR #100, opened 2022-04-17
fixing a memory leak when converting the xml to a string

scooterman:
The title says it, this change fixes a memory leak when dealing with the serialization of a document back to string.


dginev:
Thanks for patching this @scooterman !

Indeed, taking a closer look at the [xmlsave docs](https://opensource.apple.com/source/libxml2/libxml2-8/libxml2/doc/html/libxml-xmlsave.html), we've simply used the wrong method here (or omitted a call to free, either way).

PR #101, opened 2022-05-20
maybe a better approach to set library path

meltfar:
When I use this module, the rust compiler always couldn't find where the libxml2 located in. I just want to find out if there is a proper way to define a valid path of native library to compiler? 

After several hours searching, I found something in official document website that I want to discuss.

## link or links?
I tried to find a official introduction about the [`link` section](https://github.com/KWARC/rust-libxml/blob/e7e1170e31febfc6fbda12af6aac01bb23edbad3/Cargo.toml#L17), but found [nothing](https://doc.rust-lang.org/cargo/reference/manifest.html). Maybe I missed something?

Instead, there is a similar field named [`links`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-links-field) in `package` section.

By using this field, we can override the build parameters by providing a [project-level](https://stackoverflow.com/a/54801479) configuration, and no need to modify environment variables or create any softlink. 

Documents about overriding build scripts are [here](https://doc.rust-lang.org/cargo/reference/build-scripts.html#overriding-build-scripts).

In this situation, build scripts will be disabled since a target-override is provided.
> The override method would not work when we use `link` section instead of `links` field.

## conclusion
I have no idea whether this is a better approach, compared with some manual filesystem operations?

So I create this to discuss it, please feel free to comment, edit or close this PR.  

I'm not familiar with how to submit a PR, but I would be very happy if this could help somehow. And thank you for your continuously brilliant work!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,13 +9,11 @@ documentation = "http://kwarc.github.io/rust-libxml/libxml/index.html"
 readme = "README.md"
 license = "MIT"
 keywords = ["xml", "libxml","xpath", "parser", "html"]
+links = "libxml"
 build = "build.rs"
 exclude = [
   "scripts/*"
 ]
-
-[lib]
-name = "libxml"
 
 [dependencies]
 libc = "0.2"
--- a/README.md
+++ b/README.md
@@ -26,13 +26,54 @@ Before performing the usual cargo build/install steps, you need to have the rele
 On linux systems you'd need the development headers of libxml2 (e.g. `libxml2-dev` in Debian), as well as `pkg-config`.
 
 ### MacOS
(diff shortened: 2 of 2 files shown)


rudolphfroger:
The current setup isn't ideal on FreeBSD because I always need to set the `LIBXML2` environment variable. When I remove the current `panic!` from the `build.rs` file, then the library compiles just fine (if I installed `libxml2` first, of course). I don't know much about dynamic linking, but assuming the library won't be found it a bit pessimistic, in many cases it may be available at a standard location.

Both on the current `master` branch and also with this PR, it works fine with the `panic!` removed from the `build.rs`.

PR #109, opened 2023-05-02
parser: Fix computed option flag values

davidmhewitt:
`no_warning` (64) is added twice, resulting in 128 (`pedantic`), so if you have `no_warning` set, you actually end up with `pedantic` :sweat_smile: 

The rest of this code is correct.


dginev:
🤦🏻‍♂️ How did this slip in...

Thanks.

PR #116, opened 2023-07-15
Improve Error Reporting for Schema Validation

### LINKED ISSUE #115: All `StructuredError` returned by `SchemaValidationContext::validate_*` are identical
## Description
When multiple validation issues are present in a document, all of the `StructuredError` objects returned by `validate_document` or `validate_file` are the exact same object in memory.

The unit test for schema validation seems to suggest that this is expected behavior, but other FFI wrappers around `libxml` (such as: https://github.com/marudor/libxmljs2/wiki#validating-against-xsd-schema) do not exhibit this behavior. I was hoping to learn more about why this is happening and if there is any way to get each unique schema validation issue that's present.

## Reproduction
Change the following code in `tests/schema_tests.rs`, https://github.com/KWARC/rust-libxml/blob/master/tests/schema_tests.rs#L93-L98 to the following:

```rust
for err in &errors {
  println!("{:#?}", err);
  assert_eq!(
    "Element 'bad': This element is not expected. Expected is ( to ).\n",
    err.message()
  );
}
```

Then run the tests using `cargo test -- --nocapture`. In the test output, you should see something like:
```
     Running tests/schema_tests.rs (target/debug/deps/schema_tests-b2c35c81d74ebf18)

running 2 tests
StructuredError(
    0x0000000120e05748,
)
StructuredError(
    0x0000000120e05748,
)
StructuredError(
    0x0000000120e05748,
)
StructuredError(
    0x0000000120e05748,
)
StructuredError(
    0x0000000120e05748,
)
test schema_from_string ... ok
test schema_from_string_generates_errors ... ok
```
### Expected vs Actual Behavior
A unique `StructuredError` object for each unique validation issue present in the document. Instead, a Vec containing the same repeated `StructuredError` is returned.

## System Information
- cargo version: `cargo 1.70.0 (ec8a8a0ca 2023-04-25)`
- rustup show:
  - Default host: `aarch64-apple-darwin`
  - Platform: `stable-aarch64-apple-darwin (default)`
  - Compiler Version: `rustc 1.70.0 (90c541806 2023-05-31)`

JDSeiler:
I have some prior Rust experience, but I'd consider myself an intermediate Rust programmer at best, and I have very little experience with C and FFI. Point being, I'm very open to feedback on these changes.

## Motivation
Closes #115

Previously, `StructuredError` was a thin wrapper around a raw pointer to libxml2's [`xmlError` struct](https://gnome.pages.gitlab.gnome.org/libxml2/devhelp/libxml2-xmlerror.html#xmlError). This was problematic because libxml2 does not allocate separate `xmlError` structs. Instead, it uses an effectively global (or thread-global) `xmlError` that is rewritten (but does not move) every time a new error is generated. As a result, all of the errors produced by this library were all pointing to the same memory and all had the same contents: the last error libxml2 produced.

The following code from `libxml2` was consulted to confirm its behavior:
- https://github.com/GNOME/libxml2/blob/884474477284474e0151280aaa275a18e3d7a036/error.c#L470 is the function ultimately responsible for calling the structured error callback (called `schannel` in libxml)
  - line 632 is where the data is actually provided to the callback. The error object is called `to`.
  - Walking backwards, we see the error object being cleared and initialized starting on line 573
  - Walking back even farther, we see that `to` is not a fresh xmlError, but rather a reference to the `xmlLastError` global. This *either* points to a file-level `xmlError` defined in `globals.c`, or a struct-member that's part of a thread-specific global context. In either case, it does not appear that `xmlLastError` is ever allocated more than once.

## Description of Changes
I replaced the wrapper around the raw pointer with a more traditional Rust struct so that each individual error could be preserved. Because the struct no longer has any ties to the underlying C-managed data once constructed, the `Drop` implementation was removed. 

Not all of the fields in the `xmlError` struct had obvious utility, or even safe ways of managing them. The fields ommitted are:
- `str1`, `str2`, `str3` :: I couldn't find any indication as to what these would be used for. Perhaps they are "empty buckets" for users to put their own custom error messages into? It didn't seem worth spending the CPU cycles on looking at them.
- `int1` :: Similarly, I couldn't find any information on this aside from the cryptic "extra number information".
- `ctxt`, `node` :: Both of these are `void *` and I have 0 clue how to safely include them without massively complicating things.

Regarding enums, the `xmlErrorLevel` field was converted to a normal Rust enum because it's small. The `code` and `domain` fields were included in their raw forms because they might be useful to someone. However, the `xmlErrorDomain` and `xmlParserError` C enums are 30 and 700+ (!!) members respectively, so it didn't seem wise to write out Rust enum versions of them.

## Draft -> Ready for Review Checklist
- [x] :: What should the fallback value for the `XmlErrorLevel::from_raw` function be? Do we trust that `libxml2` is always going to provide something in the range `[0, 3]`, and the catch-all branch can be something like `unreachable!()`?
- [x] :: Is the type `Option<String>` acceptable for the `message` and `filename` fields? What about the lossy conversion to utf-8? I went through a few different options here. The underlying C string *could* be a null pointer (at least for `filename`) and of course there's no guarantee that the underlying C string is valid utf-8.
- [x] :: Any other finishing tasks I need to do? Run Valgrind? Improve documentation? I'm all ears.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -100,8 +100,8 @@ impl StructuredError {
   /// reporting mechanism. Instead of calling this method, you can access the 
   /// StructuredError `message` field directly.
   #[deprecated(since="0.3.3", note="Please use the `message` field directly instead.")]
-  pub fn message(&self) -> &String {
-    self.message.as_ref().unwrap()
+  pub fn message(&self) -> &str {
+    self.message.as_ref().unwrap().as_str()
   }
 
   /// Returns the provided c_str as Some(String), or None if the provided pointer is null.
--- a/src/error.rs
+++ b/src/error.rs
@@ -101,7 +101,7 @@ impl StructuredError {
   /// StructuredError `message` field directly.
   #[deprecated(since="0.3.3", note="Please use the `message` field directly instead.")]
   pub fn message(&self) -> &str {
-    self.message.as_ref().unwrap().as_str()
+    self.message.as_deref().unwrap_or("")
   }
 
   /// Returns the provided c_str as Some(String), or None if the provided pointer is null.
--- a/src/error.rs
+++ b/src/error.rs
@@ -100,10 +100,8 @@ impl StructuredError {
   /// reporting mechanism. Instead of calling this method, you can access the 
   /// StructuredError `message` field directly.
   #[deprecated(since="0.3.3", note="Please use the `message` field directly instead.")]
-  pub fn message(&self) -> String {
-    self.message
-      .clone()
-      .unwrap_or("".to_string())
+  pub fn message(&self) -> &String {
+    self.message.as_ref().unwrap()
   }
 
   /// Returns the provided c_str as Some(String), or None if the provided pointer is null.
--- a/src/error.rs
+++ b/src/error.rs
@@ -3,60 +3,97 @@
 //!
 use super::bindings;
 
-use std::ffi::CStr;
+use std::ffi::{c_char, c_int, CStr};
 
-/*
-TODO:
-The reason that the validation errors are duplicated is because libxml mutates
-the same memory location for each error. This means that every StructuredError
-object is actually wrapping the same pointer, and so every StructuredError ends
-up with the contents of the last error that was generated.
-
-My tentative plan is to rewrite StructuredError so that it contains all the same
-fields as libxml's xmlError struct, and update the plumbing so that the errors
-are copied into a new StructuredError object for each error that is generated.
-
-After rummaging around in the libxml source code, I don't think we need to be calling
-xmlResetError, at least in the new version where we'll be copying the data out into
-a normal Rust struct.
-
-In error.c, the __xmlRaiseError function appears to be what ultimately calls the
-structured error handler (defined in schema/common, in this crate). __xmlRaiseError
-does not allocate any new memory for the error it reports. Instead, it refers to a
-a global: `xmlLastError`. In a threaded context, `xmlLastError` is buried inside
-a "global state" struct that is unique per thread. Otherwise, xmlLastError is file
-level variable that is globally shared through macro magic I don't quite grasp.
-
-Point being, I don't think we need to call resetError as part of a Drop impl.
-Even in a threaded context I don't think it'd make a difference, because internally
-it calls memset instead of free, so it doesn't release any memory to the system.
-*/
-
-/// Wrapper around xmlErrorPtr
+/// Rust enum variant of libxml2's xmlErrorLevel
 #[derive(Debug)]
-pub struct StructuredError(*mut bindings::_xmlError);
+pub enum XmlErrorLevel {
+  /// No error
+  None,
+  /// A simple warning
+  Warning,
+  /// A recoverable error
+  Error,
+  /// A fatal error
+  Fatal,
+}
 
-impl StructuredError {
-  /// Wrap around and own a raw xmllib2 error structure
-  pub fn from_raw(error: *mut bindings::_xmlError) -> Self {
-    Self(error)
-  }
-
-  /// Human-readable informative error message
-  pub fn message(&self) -> &str {
-    let msg = unsafe { CStr::from_ptr((*self.0).message) };
-
-    msg.to_str().unwrap()
-  }
-
-  /// Return a raw pointer to the underlying xmlError structure
-  pub fn as_ptr(&self) -> *const bindings::_xmlError {
-    self.0 // we loose the *mut since we own it
+impl XmlErrorLevel {
+  /// Convert an xmlErrorLevel provided by libxml2 (as an integer) into a Rust enum
+  pub fn from_raw(error_level: bindings::xmlErrorLevel) -> XmlErrorLevel {
+    match error_level {
+      bindings::xmlErrorLevel_XML_ERR_NONE => XmlErrorLevel::None,
+      bindings::xmlErrorLevel_XML_ERR_WARNING => XmlErrorLevel::Warning,
+      bindings::xmlErrorLevel_XML_ERR_ERROR => XmlErrorLevel::Error,
+      bindings::xmlErrorLevel_XML_ERR_FATAL => XmlErrorLevel::Fatal,
+      _ => XmlErrorLevel::None, // TODO: What is the right fallback here?
+    }
   }
 }
 
-impl Drop for StructuredError {
-  fn drop(&mut self) {
-    unsafe { bindings::xmlResetError(self.0) }
+/// Wrapper around xmlErrorPtr.
+/// Some fields have been omitted for simplicity/safety
+#[derive(Debug)]
+pub struct StructuredError {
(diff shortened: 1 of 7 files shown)


src/error.rs
```diff
@@ -3,33 +3,97 @@
...
-impl Drop for StructuredError {
-  fn drop(&mut self) {
-    unsafe { bindings::xmlResetError(self.0) }
+  /// Returns the provided c_str as Some(String), or None if the provided pointer is null.
+  unsafe fn convert_to_owned(c_str: *mut c_char) -> Option<String> {
+    if c_str.is_null() {
+      return None;
+    }
+
+    let raw_str = CStr::from_ptr(c_str);
```
imcsk8, line 96: This is the only unsafe statement, might be more clear to have:

```
let raw_str = unsafe { CStr::from_ptr(c_str) };
```

Instead of making the whole function unsafe

imcsk8 (commented):
I've also spent some time working on a solution and came out with the same approach. I will test your patch and send you more comments if needed.

dginev:
@JDSeiler thank you for the thorough work in this PR, and thanks to @imcsk8 for the unexpected review. This is already more than the usual PR burden on this wrapper repo, I appreciate the contributions. Maybe it is a sign of the crate starting to mature.

As to the questions by @JDSeiler :

- I think `0..=3` is indeed a reliably documented range for [xmlErrorLevel](https://gnome.pages.gitlab.gnome.org/libxml2/devhelp/libxml2-xmlerror.html#xmlErrorLevel), so your suggestion to make other cases `unreachable!()` is the correct design to signal that expectation.

- Running valgrind on the updated test is a good idea, though we haven't done it consistently in the past. The usual run of
   ```
   valgrind --leak-check=full ./target/debug/deps/schema_tests-1dd74906a490bf98
   ```
   seems to indicate that there are no issues with the new test on a local build here.

- The `Option<String>` is fine in general, the crate uses that return type on occasion. But I would also honor the request by @imcsk8 and keep the original message around as [#[deprecated]](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute) from version 0.3.3. We can keep it around until 0.4.0 - although if someone wants to sink in some time to improve the crate, I am more than happy to recruit co-maintainers, and have others speak into the versioning roadmap.

- clippy also caught that `StructuredError::from_raw` should be marked as an unsafe function, see (not_unsafe_ptr_arg_deref)[https://rust-lang.github.io/rust-clippy/master/index.html#not_unsafe_ptr_arg_deref]

- I usually also update `CHANGELOG.md` in the crate root when a new PR gets merged - I can do that after merging here, or you're welcome to update it yourself if you want to choose the exact wording.

---

Aside: What actually stuck out to me is libxml's decision to add trailing `\n` newline chars to the error values. I am a bit uneasy about exposing those from the Rust struct, without first running a `.trim()` on them, especially if we are allocating a String anyway. But I could see this going either way.

src/error.rs
```diff
@@ -3,33 +3,114 @@
...
-  pub fn as_ptr(&self) -> *const bindings::_xmlError {
-    self.0 // we loose the *mut since we own it
+  /// Human-readable informative error message.
+  /// 
+  /// This function is a hold-over from the original bindings to libxml's error
+  /// reporting mechanism. Instead of calling this method, you can access the 
+  /// StructuredError `message` field directly.
+  #[deprecated(since="0.3.3", note="Please use the `message` field directly instead.")]
+  pub fn message(&self) -> &String {
+    self.message.as_ref().unwrap()
```
dginev, line 104: To keep the original `&str` type, I think this is a workable variation:
```rust
pub fn message(&self) -> &str {
  self.message.as_deref().unwrap_or("")
}
```

dginev (approved):
I left a small type suggestion, and will wait for a final nod from @imcsk8 , but we're pretty much good to merge and ship a v0.3.3 here.

Thanks again @JDSeiler

src/error.rs
JDSeiler: Ah excellent! I was about to push a change where the implementation was `self.message.as_ref().unwrap().as_str()`. I didn't think it was even possible to get *both* the `&str` return type *and* a default value working, since the compiler was quite mad about returning references to local values.

I'll push up that change shortly.

JDSeiler:
> I left a small type suggestion, and will wait for a final nod from @imcsk8 , but we're pretty much good to merge and ship a v0.3.3 here.
> 
> Thanks again @JDSeiler

Awesome, thank you (and @imcsk8 !) for feedback on the changes. This was a great learning experience!

imcsk8:
>   ... I am more than happy to recruit co-maintainers, and have others speak into the versioning roadmap.

I'll be happy to help maintaining. What do you need me to do?

dginev:
@imcsk8 basically what you just did in reviewing here, whenever you have time/interest.

I am currently available enough to ship minor releases, but I have been a little easy-going with spending time in developing the crate. Btw this is also how @triptec got added in here some time back, when he had a season of libxml work.

And since this is currently a care-free open source crate you can also just sit on the rights without doing anything, it keeps the crate safer in case I'm missing for whatever reasons.

src/error.rs
```diff
@@ -3,33 +3,114 @@
...
+    let error = *error_ptr;
+    let message = StructuredError::ptr_to_string(error.message);
+    let level = XmlErrorLevel::from_raw(error.level);
+    let filename = StructuredError::ptr_to_string(error.file);
+
+    let line = if error.line == 0 {
+      None
+    } else {
+      Some(error.line)
+    };
```
imcsk8, line 79: This might be too much but I leave you a more idiomatic way of doing these assignments

```suggestion
    let line = match error.line {
      0 => None,
      _ => Some(error.line)
    };
```

dginev:
@imcsk8 and in the interest of speed, I just added you as an admin to the repo (you should get an invite).

Feel free to merge this PR in as a warm-up, if the general setup sounds reasonable enough.

src/error.rs
```diff
@@ -3,33 +3,114 @@
...
+    let col = if error.int2 == 0 {
+      None
+    } else {
+      Some(error.int2)
+    };
```
imcsk8, line 84:
```suggestion
    let col = match error.int2 {
      0 => None,
      _ => Some(error.int2)
    };
```

imcsk8 (approved):
LGTM

I approved but also left a couple of suggestions in case you want to add them to the PR

src/error.rs
dginev: Well, the `if`-`else` variant is the first snippet in the [std::option](https://doc.rust-lang.org/std/option/) doc page, so it's not really criminal... This one may come down to taste.

imcsk8: > Well, the `if`-`else` variant is the first snippet in the [std::option](https://doc.rust-lang.org/std/option/) doc page, so it's not really criminal... This one may come down to taste.

You're right, it's a matter of taste :+1: 

Thanks @JDSeiler for the patch and thanks @dginev  for the quick review and merge!!

I'm using heavily this library and I would like to help on whatever I can.

Cheers!

Issue #125, opened 2024-02-23
Missing the ability to get an attribute with no namespace


anwaralameddin:
Using the current `Node` API, it does not seem always possible to retrieve the value of a property with no namespace when multiple properties have the same local name.

The method `get_property` returns the value of the first property with a given local name, regardless of namespace; it cannot retrieve the value of the nonprefixed property `bar` in the node below when `ns:bar` and `bar` appear in this order:
```XML
<foo xmlns:ns="http://example.com" ns:bar="NS BAR" bar="bar" />
```

The library already provides two methods, `get_property_ns` and `get_property`, that utilise `xmlGetNsProp` and `xmlGetProp`, respectively. I think I need a third, say `get_property_no_ns`, that similarly utilises `xmlGetNoNsProp`.

Please let me know if this can already be achieved differently.

dginev:
I think you are making a very reasonable suggestion, and `get_property_no_ns` continues the existing pattern.

If you feel like making a PR, I'm happy to accept it.

anwaralameddin:
Thank you! While writing the test, I noticed the same issue for `remove_attribute`. That, it removes the first property regardless of namespace. However, it is unclear to me whether the needed bindings exist to address this. Anyway, It could be better handled separately.

PR #127, opened 2024-02-24
Add get_property_no_ns to Node and RoNode

anwaralameddin:
The method `get_property_no_ns` was added to `Node` and `RoNode` in line with this [issue](https://github.com/KWARC/rust-libxml/issues/125).

The method `get_property` returns the value of the first property with a given local name, regardless of namespace. This can lead to undesirable results when an element has multiple properties with the same local name. For example, calling it with "bar" on the node below returns "NS BAR" instead of "BAR".
```XML
<foo xmlns:ns="http://example.com" ns:bar="NS BAR" bar="BAR" />
```
On the other hand, the method `get_property_no_ns` returns the value of the nonprefixed property.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [0.3.4] (in development)
 
+### Added
+
+* Node methods: `get_property_no_ns` (alias: `get_attribute_no_ns`)
+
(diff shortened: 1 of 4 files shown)


dginev (approved):
Perfect, this is completely in line with the existing methods. Thank you!

Issue #104, opened 2022-11-29
missing ability to get all attributes including namespace


faassen:
Consider this:

```xml
<foo xmlns:ns="http://example.com" bar="BAR" ns:bar="NS BAR" />
```

How can I find out that attributes `bar` and `ns:bar` both exist?

The `Node` API offers `get_attributes` which returns a  `HashMap` with all attributes, but this loses namespace information and I only get `bar` once.

I can access individual attributes by namespace using `get_attribute_ns`, but I have no way to find out which attributes exist in the first place. I would expect something like `get_attributes_ns` which returns a vec of `Node` (of type `AttributeNode`, like what is returned by `get_attribute_node`) in order to find this out. Or alternatively a `get_attribute_names_ns` which gives a vec of attribute name, attribute namespace tuples.

Am I missing something or this is something the API doesn't implement yet?

dginev:
> Am I missing something or this is something the API doesn't implement yet?

I think the namespace support is lacking in the current wrapper, so this would need a new method, likely `get_attributes_ns`, as you suggest.

anwaralameddin:
I needed the same and have implemented it with the signature
```rust
pub fn get_properties_ns(&self) -> HashMap<(Option<Namespace>, String), String>
```
For my use case, I have implemented `PartialEq` and `Hash` for `Namespace` using the methods `get_prefix` and `get_href` rather than `ns_ptr`. 

Shall I create a pull request? Or do you have a different signature or implementation in mind?

dginev:
@anwaralameddin your suggestion sounds reasonable to me. I'm happy to accept such a PR!

PR #128, opened 2024-02-24
Add get_properties_ns to Node and RoNode

anwaralameddin:
The method `get_properties_ns` was added to `Node` and `RoNode` in line with this [issue](https://github.com/KWARC/rust-libxml/issues/104).

It returns a `HashMap` of attribute values, which is indexed by tuples of the attributes and their namespaces.

The code is similar to that of `get_properties`. However, `get_properties` used unnecessary vector allocation `attr_names`, which is removed in the PR.

The indexing mentioned above required implementing the traits `Hash`, `PartialEq` and `Eq` for `Namespace`. These were implemented using the methods `get_prefix ` and `get_href` instead of `ns_ptr`. This choice does not affect the method `get_properties_ns` as a prefix is bound to a single namespace name in a given scope. However, this has the effect that the two namespaces declared below are considered identical.
```XML
<root>
    <child1 xmlns:foo="http://www.example.com/myns" attribute="value1" foo:attribute="foo1" />
    <child2 xmlns:foo="http://www.example.com/myns" attribute="value2" foo:attribute="foo2" />
</root>
```

### DIFF
--- a/src/tree/namespace.rs
+++ b/src/tree/namespace.rs
@@ -2,6 +2,7 @@
 //!
 use std::error::Error;
 use std::ffi::{CStr, CString};
+use std::hash::{Hash, Hasher};
 use std::ptr;
 use std::str;
 
@@ -14,6 +15,21 @@ use crate::tree::node::Node;
 pub struct Namespace {
   ///libxml's xmlNsPtr
   pub(crate) ns_ptr: xmlNsPtr,
+}
+
+impl PartialEq for Namespace {
+  fn eq(&self, other: &Self) -> bool {
+    self.get_prefix() == other.get_prefix() && self.get_href() == other.get_href()
+  }
+}
+
+impl Eq for Namespace {}
+
+impl Hash for Namespace {
+  fn hash<H: Hasher>(&self, state: &mut H) {
+    self.get_prefix().hash(state);
+    self.get_href().hash(state);
+  }
 }
 
 impl Namespace {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@
 
 ### Added
 
-* Node methods: `get_property_no_ns` (alias: `get_attribute_no_ns`)
+* Node methods: `get_property_no_ns` (alias: `get_attribute_no_ns`), `get_properties_ns` (alias: `get_attributes_ns`) 
+* Added implementations of `Hash`, `PartialEq` and `Eq` traits for `Namespace` 
 
 ## [0.3.3] 2023-17-07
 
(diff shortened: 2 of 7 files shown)


src/tree/namespace.rs
```diff
@@ -16,6 +17,21 @@ pub struct Namespace {
   pub(crate) ns_ptr: xmlNsPtr,
 }
 
+impl PartialEq for Namespace {
+  fn eq(&self, other: &Self) -> bool {
+    self.get_prefix() == other.get_prefix() && self.get_href() == other.get_href()
+  }
+}
```
dginev, line 24: This is a little tricky, since it can be surprising depending on what expectation someone approaches the `Namespace` struct.

Internally, Namespace stores a libxml2 `xmlNsPtr`. So one could have expected pointer equality. Alternatively, one could have expected the equality to be in the XML sense - if the URL is identical, then the namespaces are equal, even if the prefixes are different.

So introducing the "deepest" kind of equality here (same prefix and same namespace) isn't immediately expected behavior. Additionally, it is the slowest variant performance-wise, and uses dedicated allocations.

Hm... I am not opposed to merging it as introduced here, but I wonder if we'll get requests to modify it down the road. For now a clarifying comment on the Namespace struct could help.

src/tree/namespace.rs
```diff
@@ -16,6 +17,21 @@ pub struct Namespace {
...
+
+impl Eq for Namespace {}
+
+impl Hash for Namespace {
+  fn hash<H: Hasher>(&self, state: &mut H) {
+    self.get_prefix().hash(state);
+    self.get_href().hash(state);
+  }
+}
```
dginev, line 33: Similarly to above, using the pointer directly could be another reasonable approach to hashing.

dginev (commented):
Thank you for the contribution, and the new tests!

Everything looks great, but I am wondering if the equality implementation would cause confusion down the road. Not opposed to merging, but wondering if you have additional motivation to share.

anwaralameddin:
Thank you for your review and comments!

I actually had conflicting thoughts about the three options. As you say, it depends on what one expects from the `Namespace` structure.

I would've liked it to be true to the specification, but I rolled out the URI comparison because I don't see it as an implementation of the XML namespace. I think of it as a `NamespaceBinding` structure, binding prefixes to namespace names, which is one reason for the current choice. Also, I wanted to preserve the (parts of) documents that are parsed and written out without further processing and didn't want this to potentially change prefixes.

Using the pointer felt unnecessarily too strict, given the usage of namespaces, but I don't have a strong argument against it. If you wish, we can change it to the strictest equality (using the `xmlNsPtr`), and those who desire a more relaxed notion can group the results as needed.

PR #129, opened 2024-02-25
Add no_ns variants for Node and RoNode methods

anwaralameddin:
This completes what started in another [PR](https://github.com/KWARC/rust-libxml/pull/127) by adding the node methods `has_property_no_ns`, `remove_property_no_ns` and `get_property_node_no_ns`. These methods target nonprefixed properties and are needed when an element has multiple properties with the same local name, as in the example below:
```XML
<foo xmlns:ns="http://example.com" ns:bar="NS BAR" bar="BAR" />
```
Also, the method `get_property_node_ns` was added.

There was no added functionality to be gained by introducing `set_property_no_ns` as the existing method `set_property` sets the value for an attribute with no namespace.

While this adds functionality, it seems too verbose and repetitive. More ideally, one should have two variants of each method. One only takes an attribute name and does not consider namespaces:
```rs
pub fn foo_property(&self, name: &str)
``` 
The other takes an attribute name and, optionally, a namespace name:
```rs
pub fn foo_property_ns(&self, name: &str, ns: Option<&str>)
``` 
However, this will introduce braking changes for `has_property_ns`, `get_property_ns` and `remove_property_ns`. Could this be considered now?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Added
 
-* Node methods: `get_property_no_ns` (alias: `get_attribute_no_ns`), `get_properties_ns` (alias: `get_attributes_ns`) 
+* Node methods: `get_property_no_ns` (alias: `get_attribute_no_ns`), `get_properties_ns` (alias: `get_attributes_ns`), `has_property_no_ns` (alias: `has_attribute_no_ns`), `remove_property_no_ns` (alias: `remove_attribute_no_ns`), `get_property_node_ns` (alias: `get_attribute_node_ns`), `get_property_node_no_ns` (alias: `get_attribute_node_no_ns`)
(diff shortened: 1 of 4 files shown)


dginev:
XML namespaces are notorious for causing API pain. There isn't a perfect solution for the current design of XML, as far as I'm concerned.

The situation you are guarding for should be an exceptionally rare one - and these verbose repetitive methods are the price one pays for wading into a design that shares attribute names in differing namespaces.

The simple methods address the 99% of use cases that have unique local names cross-namespace, and I don't think we should be breaking them. If anything, I would personally be discouraging people that have nodes with `foo="val" ns1:foo="val1" ns2:foo="val2"`, although I am certain one can motivate using these (such as simplifying XSLT for different targets by focusing on a different namespace for each target format).

I recall Perl has addressed this by adding a runtime cost where the namespace is checked in each getter/setter. But I would prefer redundancy to runtime penalty in Rust.

PR #134, opened 2024-05-15
support windows mingw64 gnu

netoffice-zz:
current rust-libxml isn't support mingw64 develop environment, this pull request  realize it, i has tested

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,8 +20,11 @@ name = "libxml"
 [dependencies]
 libc = "0.2"
 
-[target.'cfg(windows)'.build-dependencies]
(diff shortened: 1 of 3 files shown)


dginev:
Sorry for the slow response here. Could you give me an OS example where this was needed? Also, do you think we can add it to github actions CI to ensure it's working as expected?

I am not too familiar with those systems.

dginev:
@netoffice-zz it took a while to refocus on windows installations for rust-libxml - do you have time+interest to sync up this PR with the latest repository master?

We just got CI in for Windows in #156, but only with the msvc target. If you'd like to extend the windows CI target with gnu to make sure that target keeps passing, I'd be happy to merge.

If not, let me know, I can try that as well.

Issue #133, opened 2024-05-08
How to receive errors from xpath queries.


hurricane-socrates:
I have an xpath query involving namespaces that executes successfully in xmllint. The query does not execute using rust-libxml. The Result Err is, as you know, ()

This makes debugging quite difficult.

How does one debug xpath queries using rust-libxml?
```
        if let Ok(context) = Context::new(&doc) {
            // let ns = get_default_namespace(&mut root);
            context.register_namespace("scxml", "http://www.w3.org/2005/07/scxml").unwrap();
            println!(
                "node count: {}",
                root
                    .findnodes(".//scxml:initial/scxml:transition[@target]")
                    .unwrap()
                    .len()
            );
```
Gives the following error
```
thread 'scxml::test_scxml::test_enter_states_010' panicked at src/scxml.rs:1101:22:
called `Result::unwrap()` on an `Err` value: ()
```
The xmllint session
```
xmllint --shell /tmp/foo.xml
/ > setns scxml=http://www.w3.org/2005/07/scxml 
/ > du .//scxml:initial/scxml:transition[@target]
scxml:transition
```
foo.xml
```

<?xml version="1.0" encoding="UTF-8"?>
<scxml:scxml xmlns:scxml="http://www.w3.org/2005/07/scxml" binding="late" version="1.0">
            <state id="xyzzy">
                <initial id="elmer"/>
                <state id="elmer">
                    <state id="charlie">
                        <initial id="frank"/>
                        <state id="frank"/>
                    </state>
                </state>
            </state>
            <state id="fred">
                <initial id="fubar"/>
                <state id="fubar"/>
            </state>
        <scxml:initial><scxml:transition target="xyzzy"/></scxml:initial></scxml:scxml>

**Comment by dginev** [comment]
It should be possible to pass through the error information to the Rust layer. Would you like to work on a PR for that @hurricane-socrates ? It ought to be similar in spirit to #116

**Comment by hurricane-socrates** [comment]
Sure. I'll take a look at that PR

**Comment by hurricane-socrates** [comment]
I could use some help locating the routine ```xmlXPathSetErrorHandler```
AFAICT, this is the routine signature to declare in bindings.rs
```
extern "C" {
  pub fn xmlXPathSetErrorHandler(ctxt: xmlXPathContextPtr, handler: xmlStructuredErrorFunc, data: *mut ::std::os::raw::c_void,);
}
```
[...]

PR #135, opened 2024-05-20
Implement xpath error message handling

hurricane-socrates:
See #133 for context.

Two issues so far: 
1. The first test isn't correct yet. I wanted to start this PR. 
2. After each source change I have to redefine `LIBXML2`. I'm not sure how to tell Cargo where the correct libxml2 resides. I think it's finding the Apple version, not the `brew` version. There's also a MacPorts version that might be in the search path.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 *.rlib
 *.dll
 Cargo.lock
+.DS_Store
 
 # Executables
 *.exe
--- a/src/bindings.rs
+++ b/src/bindings.rs
@@ -14927,6 +14927,13 @@ extern "C" {
     -> ::std::os::raw::c_int;
 }
 extern "C" {
+  pub fn xmlXPathSetErrorHandler(
+    ctxt: xmlXPathContextPtr,
+    handler: xmlStructuredErrorFunc,
+    data: *mut ::std::os::raw::c_void,
+  );
+}
+extern "C" {
   pub fn xmlXPathNodeEval(
     node: xmlNodePtr,
     str: *const xmlChar,
--- a/src/schemas/mod.rs
+++ b/src/schemas/mod.rs
@@ -16,3 +16,4 @@ use schema::Schema; // internally handled by SchemaValidationContext
 
 pub use parser::SchemaParserContext;
 pub use validation::SchemaValidationContext;
+pub use common::structured_error_handler;
--- a/src/xpath.rs
+++ b/src/xpath.rs
@@ -1,9 +1,13 @@
 //! The `XPath` functionality
 
-use crate::bindings::*;
-use crate::c_helpers::*;
-use crate::readonly::RoNode;
-use crate::tree::{Document, DocumentRef, DocumentWeak, Node};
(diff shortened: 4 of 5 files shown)


anwaralameddin:
On macOS, [pkg-config-0.3.2](https://github.com/KWARC/rust-libxml/blob/0dcce40488fd3a3078b4aa2a0ce06a94372c916b/Cargo.toml#L26-L27) is used in the [build script](https://github.com/KWARC/rust-libxml/blob/0dcce40488fd3a3078b4aa2a0ce06a94372c916b/build.rs#L49-L57) to locate <code>libxml2</code> when <code>LIBXML2</code> is not set. Setting the [<code>atleast_version</code>](https://docs.rs/pkg-config/0.3.2/pkg_config/struct.Config.html#method.atleast_version) and replacing the function [<code>find_library</code>](https://docs.rs/pkg-config/0.3.2/pkg_config/fn.find_library.html) with the method [<code>find</code>](https://docs.rs/pkg-config/0.3.2/pkg_config/struct.Config.html#method.find) might help, but I have not tested it. For local tests, it might be simpler to reorder the entries in the library search paths.

Anyway, this breaks tests for anyone using a released version of <code>libxml2</code> regardless of OS. Can the PR objective be achieved using <code>xmlSetStructuredErrorFunc</code> instead of <code>xmlXPathSetErrorHandler</code>? If so, you could consider using it until version 2.13.0 is resealed and widely adopted. For reference, ubuntu's latest release still uses <code>libxml2</code> version [2.9.14](https://pkgs.org/search/?q=libxml2-dev). Otherwise, if <code>xmlSetStructuredErrorFunc</code> does not work here, you could consider [conditional compilation](https://doc.rust-lang.org/reference/conditional-compilation.html) on the <code>libxml2</code> version.

hurricane-socrates:
Yeah, I was going around on the question of version support. I have no problem waiting. I see Debian Stable is also on the 2.98 branch

I couldn't get the build script to work with the updated HEAD version. I can see it uses pkg-config, and brew creates a correct definition. Nevertheless... I'll try pounding on it some more. It's probably got something to do with brew not symlinking libxml2 b/c Apple

In the meantime, the whole point of this task for me was that libxml2 reported line & col. It has stopped doing that reporting. Very annoying.

I'll look at  `xmlSetStructuredErrorFunc`. It doesn't ring a bell right now. I'm sure it'll be as useful as `xmlXPathSetErrorHandler` has turned out to be. Right now I am annoyed at developing OSS on macos. At least three versions of libxml: Apple's, MacPort's, Brew's.

hurricane-socrates:
I'd like to withdraw this PR.
0) Implementing ```xmlXPathSetErrorHandler``` for this task increases this project's technical debt;
1) libxml has stopped reporting detailed xpath errors. I'll look into why this is happening. The change nullifies my reasons for using ```xmlSetStructuredErrorFunc```.

dginev:
Thank you for investigating this @hurricane-socrates and feel free to close here if you've chosen to do so.

PR #140, opened 2025-02-27
tests: avoid writing to the source directory of the crate

wetneb:
This makes for a better separation between source code and test execution data. It is needed to make tests pass in a Debian build environment where the source directory is not writeable.


dginev:
Thank you, sounds healthy.

PR #150, opened 2025-04-21
Compatibility with libxml2 >= 2.12

### LINKED ISSUE #147: The new version causes build issues
Hello there,

I'm building a musl based rust build container with some pre-compiled C libraries, including libxml2.
All previous versions seem to work perfectly and build without issues.

But since the release of the new version this now breaks.
Not specifically sure why this is happening though, as i have not yet look into it deeply.
But maybe somebody here knows a way to fix it, or see what the issues more quickly :) 

The error can be viewed here:
https://github.com/BlackDex/rust-musl/actions/runs/14512752243/job/40714970804#step:33:324

Any pointers would be great.

wetneb:
Fixes #147.

First, opening the PR just with the added CI job to demonstrate its failure (see [the logs](https://github.com/KWARC/rust-libxml/actions/runs/14570632785/job/40867153540)).
Then, fixing the build by detecting the libxml2 version via pkg-config and adapting the code accordingly.

Note that in my previous PR I generated the `default_bindings.rs` with libxml 2.9.14. We can also revisit this choice later, possibly introducing different default bindings for < 2.12 and for >= 2.12, letting the user select which one via some environment variable. It would also be good to introduce a similar auto-detection logic on Windows via `vcpkg`.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -27,3 +27,35 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
+
+  test-libxml2_12:
+    name: With libxml 2.12
+    runs-on: ubuntu-latest
+    steps:
+      - name: install dependencies
+        uses: ryankurte/action-apt@v0.2.0
+        with:
+          packages: "libpython3-dev"
+      - uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


dginev:
Great diagnostgic and PR, much appreciated. Will push out v0.3.5 now.

PR #151, opened 2025-04-21
Support for libxml 2.13

wetneb:
I couldn't figure out how to get rid of the remaining test errors, so leaving this as a draft, if anyone wants to pick this up from there.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -28,8 +28,11 @@ jobs:
         with:
           command: test
 
-  test-libxml2_12:
-    name: With libxml 2.12
+  test-newer-libxml2:
+    strategy:
+      matrix:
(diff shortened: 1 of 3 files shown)


wetneb:
Since the addition of the semicolon works for @BlackDex, I've changed this PR to just add support for libxml 2.13, not 2.14 (as there seem to be more test failures for this one).

dginev:
It appears that for 2.13.8 the tests may also encounter a test condition leading to a segfault: 
https://github.com/KWARC/rust-libxml/actions/runs/14591808764/job/40928557349

I'll take a closer look, could have surfaced a Rust issue.

BlackDex:
Maybe using miri could help find the specific issue?
Or there are still other bindgen's a bit different?

PR #155, opened 2025-04-23
Patches for libxml >=2.12

### LINKED ISSUE #107: Cargo test locking up
I have come across an issue with cargo locking up, I expect due to some race condition? The following code exhibits the behavior on my local development machine and on our CI servers:

``` rust
use libxml::parser::Parser;
use libxml::schemas::{SchemaParserContext, SchemaValidationContext};

fn main() {
    println!("Hello, world!");
}

pub fn do_xml_things(schema: &str, xml: &str) {
    let mut schema_parser = SchemaParserContext::from_buffer(schema);
    let mut xsd = SchemaValidationContext::from_parser(&mut schema_parser).unwrap();

    let doc = Parser::default().parse_string(xml).unwrap();
    xsd.validate_document(&doc).unwrap();
}

#[cfg(test)]
mod test {

    use super::*;

    const SCHEMA: &str = r#"<?xml version="1.0"?>
    <xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
    
    <xs:element name="note">
      <xs:complexType>
        <xs:sequence>
          <xs:element name="to" type="xs:string"/>
          <xs:element name="from" type="xs:string"/>
          <xs:element name="heading" type="xs:string"/>
          <xs:element name="body" type="xs:string"/>
        </xs:sequence>
      </xs:complexType>
    </xs:element>
    
    </xs:schema>"#;

    #[test]
    fn test_a() {
        do_xml_things(
            SCHEMA,
            "<note><to>Anyone</to><from>Me</from><heading /><body /></note>",
        );
    }

    #[test]
    fn test_b() {
        do_xml_things(
            SCHEMA,
            "<note><to>Anyone</to><from>Me</from><heading /><body /></note>",
        );
    }

    #[test]
    fn test_c() {
        do_xml_things(
            SCHEMA,
            "<note><to>Anyone</to><from>Me</from><heading /><body /></note>",
        );
    }

    #[test]
    fn test_d() {
        do_xml_things(
            SCHEMA,
            "<note><to>Anyone</to><from>Me</from><heading /><body /></note>",
        );
    }

    #[test]
    fn test_e() {
        do_xml_things(
            SCHEMA,
            "<note><to>Anyone</to><from>Me</from><heading /><body /></note>",
        );
    }
}
```

Running ``cargo test`` sometimes (1 in 3 times for me) results in all the tests locking up. Running ``RUST_TEST_THREADS=1 cargo test`` removes the issue altogether.

I will implement setting the tests as a workaround, but I am not sure if this is an underlying libxml2 issue, or something specific to the crate.

Thoughts?

### LINKED ISSUE #152: Building issue with libxml 2.14.1 : undefined symbol: initGenericErrorDefaultFunc
LibXML 2.14.1 removed `initGenericErrorDefaultFunc` from the exported symbols causing a linking error during build:

`ld: error: undefined symbol: initGenericErrorDefaultFunc`

The error was spotted when building [hurl](https://github.com/Orange-OpenSource/hurl/tree/master), who uses rust-libxml 0.3.3 , but the issue will probably the same with the latest version of rust-libxml.

In rust-libxml 0.3.3, I did a quick fix by removing `libxml::bindings::initGenericErrorDefaultFunc` and `silentErrorFunc` from src/bindings.rs,. 
This solves the build issue with libxml 2.14.1 without visible regression with  libxml 2.11.9 . Removing the C implementation of `silentErrorFunc` it's probably also required.

### LINKED ISSUE #153: Schema test failure in newer libxml versions
I got a little closer to a recent test failure, now that we have newer libxml versions in CI and do dynamic builds with bindgen (since v0.3.4). It gets logged as:

```
thread 'schema_from_string_reports_unique_errors' panicked at tests/schema_tests.rs:169:5:
Failed to parse schema with 0 errors
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Internal error: xmlSchemaPValAttrNodeValue, validation using the given type is not supported while parsing a schema.
```

From the run at:

https://github.com/KWARC/rust-libxml/actions/runs/14594168438/job/40936002478#step:6:216

I am not sure if this is the concrete issue however, as the failures are rather transient - the test needs to run on a libxml2 >=2.12 and there need to be at least two tests active in `schema_tests.rs` to encounter a failure.

dginev:
Fixes #153 . Fixes #107. Fixes #152 .

I am not too happy with this PR suggestion. I rearranged the schema validation test to ensure it runs single-threaded, since isolating the exact cause for the segfault wasn't sufficiently clear. I would have preferred keeping the test as-is and finding the global state to guard in libxml.

The schema validation work was contributed by @cbarber so if they have time to take a look at some later point, they're welcome to.

Related in theme, but not in actual code path, I added the `.call_once` idea from #107 , which was guarding this exact kind of issue in the main XML parser.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -31,7 +31,7 @@ jobs:
   test-newer-libxml2:
     strategy:
       matrix:
-        libxml_version: ["2.12.9", "2.13.8"]
+        libxml_version: ["2.12.9", "2.13.8","2.14.1"]
     name: With libxml ${{ matrix.libxml_version }}
     runs-on: ubuntu-latest
     steps:
@@ -59,4 +59,3 @@ jobs:
           command: test
         env:
           LD_LIBRARY_PATH: /usr/local/lib
-
--- a/src/default_bindings.rs
+++ b/src/default_bindings.rs
@@ -3791,9 +3791,7 @@ pub type xmlStructuredErrorFunc = ::std::option::Option<
 unsafe extern "C" {
   pub fn xmlSetGenericErrorFunc(ctx: *mut ::std::os::raw::c_void, handler: xmlGenericErrorFunc);
 }
-unsafe extern "C" {
-  pub fn initGenericErrorDefaultFunc(handler: *mut xmlGenericErrorFunc);
-}
+
 unsafe extern "C" {
   pub fn xmlSetStructuredErrorFunc(
     ctx: *mut ::std::os::raw::c_void,
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -15,6 +15,9 @@ use std::os::raw::{c_char, c_int};
 use std::ptr;
 use std::slice;
 use std::str;
+use std::sync::Once;
+
+static INIT_LIBXML_PARSER: Once = Once::new();
 
 enum XmlParserOption {
   Recover = 1,
@@ -217,6 +220,10 @@ pub struct Parser {
 impl Default for Parser {
   /// Create a parser for XML documents
   fn default() -> Self {
+    // avoid deadlocks from using multiple parsers
+    INIT_LIBXML_PARSER.call_once(|| unsafe {
+      crate::bindings::xmlInitParser();
+    });
     Parser {
       format: ParseFormat::XML,
     }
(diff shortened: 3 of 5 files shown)


wetneb:
Looks very sensible to me, although I have too little background knowledge on libxml to contribute an informed judgment.

dginev:
I'll take that 👍🏻

Btw, I now also understand the missing work from the Windows installation side. Once that is in place, will try to stabilize a v0.3.5

PR #157, opened 2025-04-28
Support Mingw-w64, also with CI target

dginev:
This PR is an update and transfer of #134 

It was much harder to stabilize than I anticipated, but this should be good enough for v0.3.5.

If actual Windows developers are reading this at a later date, please feel welcome and encouraged to contribute further improvements to the CI recipe, as well as to the approach for dealing with xmlFree. I don't develop on Windows at all, so this PR was my best effort as an outsider.

The key observation that touches the actual wrapper code is that a few of the bindings-near code paths were not sufficiently portable to work in MinGW, hence they get gradually refactored.

Coming up with a CI recipe was quite difficult, but that was mostly due to my having no experience working in either Windows environment.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -4,13 +4,14 @@ name: CI Windows
 
 jobs:
   test-default-windows:
-    name: Windows default
+    name: Windows vcpkg (default)
     runs-on: windows-latest
     env: 
       VCPKGRS_DYNAMIC: 1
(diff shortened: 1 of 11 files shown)


dginev:
I can't quite nail down what vcpkg settings I need in CI to also allow using `xmlFree` directly in that environment. Tried a range of configurations, but I am missing some details about visible env paths and linker expectations.

PR #158, opened 2025-04-28
Protect schema parsing with a mutex.

TreyE:
`xmlSchemaParse` is NOT thread-safe - it calls `xmlSchemaInitTypes` - which manipulates shared memory in a thread-unsafe way.

Another possible solution is to put some sort of thread-safe OnceCell around `xmlSchemaInitTypes` and invoke it ourselves in a thread-safe manner the first time `from_parser` is called.

An alternative 'init-once' approach has been submitted as PR #159.

### DIFF
--- a/src/schemas/schema.rs
+++ b/src/schemas/schema.rs
@@ -1,11 +1,15 @@
 //!
 //! Wrapping of the Schema (xmlSchema)
 //!
+use std::sync::Mutex;
+
 use super::SchemaParserContext;
 
 use crate::bindings;
 
 use crate::error::StructuredError;
+
+static SCHEMA_MUTEX: Mutex<()> = Mutex::new(());
 
 /// Wrapper on xmlSchema
 pub struct Schema(*mut bindings::_xmlSchema);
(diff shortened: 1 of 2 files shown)


dginev:
@TreyE can you show that this works by undoing [my change to schema_tests.rs](https://github.com/KWARC/rust-libxml/pull/155/files#diff-68a3190fe68313d4c29a5215eb56e476570443c3d6eaf16af5a11bce32681364) and getting CI to succeed on them in parallel?

I wonder if there are other places to guard besides this one to get all schema-related workflows working safely.

dginev:
The "init once" approach is something we [use already for xmlInitParser](https://github.com/KWARC/rust-libxml/blob/dc2c7c3ba3d9a3d4136f8ace6b1fb105bc62274f/src/parser.rs#L223-L226), so there is healthy precedent in that direction.

TreyE:
> The "init once" approach is something we [use already for xmlInitParser](https://github.com/KWARC/rust-libxml/blob/dc2c7c3ba3d9a3d4136f8ace6b1fb105bc62274f/src/parser.rs#L223-L226), so there is healthy precedent in that direction.

I'm about to submit the other so you have alternatives - if that's your prefered method we could examine either - or maybe roll them into some sort of global 'run-once' so we can centralize the cost - instead of checking every time one of our poor downstream clients wants to parse a schema.

TreyE:
> @TreyE can you show that this works by undoing [my change to schema_tests.rs](https://github.com/KWARC/rust-libxml/pull/155/files#diff-68a3190fe68313d4c29a5215eb56e476570443c3d6eaf16af5a11bce32681364) and getting CI to succeed on them in parallel?
> 
> I wonder if there are other places to guard besides this one to get all schema-related workflows working safely.

Yes - let me take a look.  I'll try to put this in both submissions.  Again - trying to make this "dealer's choice".

dginev:
I was reminding myself of the previous discussions around thread-safety and mutexes for this wrapper. Basically there was NO thread-safety implied, and especially not when there is mutable behavior.

A small writeup of the last time I was studying the Mutex-adjacent interfaces [is in this comment](https://github.com/KWARC/rust-libxml/issues/46#issuecomment-483278430).

To allow for thread-safe "read only" operations, we ended up with the [readonly module](https://kwarc.github.io/rust-libxml/libxml/readonly/index.html) and `RoNode`.

So I am remembering my preference to keep away from adding any new Mutex use inside this crate, in favor of adding more documentation that there is NO thread-safety, and users of anything mutable need to consider guarding their designs appropriately. It avoids creating a false perception.

TreyE:
Closing as the alternative PR #159, was accepted.

PR #159, opened 2025-04-28
Protect `xmlSchemaInitTypes` from thread-unsafe behaviour.

TreyE:
`xmlSchemaParse` calls `xmlSchemaInitTypes`.
`xmlSchemaInitTypes` is a lazy function which is only intended to be called once for optimization purposes - but libxml2 doesn't do this in a thread-safe manner.  We wrap the call in a OnceLock so that it only ever needs to be invoked once - and will do it in a thread-safe way.

This is the alternative approach to PR #158.


TreyE:
Looks like this approach is more sensitive to particular versions of LibXML (as a dumb, off the top of the head guess) - let me find out when what version was introduced.

dginev (approved):
Excellent triage and fix - much appreciated!

Issue #76, opened 2020-06-08
Move bindgen bindings into separate libxml-sys crate


lweberk:
So downstream crates that enhance the rust-libxml wrapper link against the same bindings instead of parallel ones, and eventually enable libxml2 vendoring.

Would benefit siblings like xmlsec greatly, since it allows it to link against the C library through the same bindings.

What are your thoughts about this?

dginev:
Ties into #73 of which I was recently made aware of. I have a general take that I am quite happy to merge PRs with such upgrades, but that I likely won't have the time to properly develop and test on several architectures, which is what one needs to do to make sure the multiple bindings work as advertised.

If you think the complexity of shipping multiple bindings warrants a standalone crate, I would be tempted to believe that. Can still be hosted in this repository, though if someone feels strongly about maintaining the bindings completely separately from the rust-libxml repository, I am open to that as well.

Strongly in agreement that supporting multiple architectures should happen, just not in a position to contribute it myself in the immediate future.

lweberk:
Our main angle is making dependence on libxml easier for xmlsec, by sharing the underlying API and thus linking through the same interface. That decouples xmlsec's version from the libxml as far as the low level stuff goes, and leaves only the pure Rust stuff to handle. The former would result in linkage errors, the latter in compilation errors. The second option is way safer and easier to handle.

That said, –at first glance– it would also be beneficial for #73 . 

As for implementation, let me look around and learn a little from other crates that do this (there are plenty), specially the ones that do library vendoring across platforms. Regardless of what we manage to come up with at our end, we would also not go as far as to start implementing cross-platform support. Its not our particular itch and we are pressed on resources as it is.

lweberk:
Is there any source of the headers you are currently including for the bindings generation?

lweberk:
Nevermind... missed it since it is still a PR

dginev:
Right, sorry about that - the PR contains the recipe for running bindgen, but simply running the code there isn't complete, as there is also some cleanup needed to restore the repository back to a state where no "helpers" exist in the auto-generated bindings.rs file.

MaxKingPor:
@dginev Hello, I made an example of [libxml-sys](https://github.com/MaxKingPor/rust-libxml-sys). Can it be used as a reference?

PR #166, opened 2025-05-04
add more libxml2 headers for bindgen

### LINKED ISSUE #165: Increase wrapper.h coverage (breakage with v0.3.4 and xpointer.h)
It looks like some C bindings that were available in v0.3.3 are no longer available in versions v0.3.4+, likely due to the switch from including bindings as static files in the repo to generating bindings at build time (tested with libxml version 2.13.6 on macOS).

One example:  the binding for `xmlXPtrNewContext` can be seen [here from the v0.3.3 tag](https://github.com/KWARC/rust-libxml/blob/7d8d83a66b656e2033c22a63c1dc4cd329ec9031/src/bindings.rs#L21002-L21007).

This, and the `xmlXPtrEval` binding is used by the `samael` crate when using the `xmlsec` feature. `samael` pinned `libxml` to the semver range of `^0.3.1`, and so this regression is allowed by semver resolution rules, but introduces a breaking change that leads to a compile-time error.

I believe the 0.3.4 release should have been 0.4.0 to reflect this breaking change to the public API of this crate. 

See [this issue in the `samael` project](https://github.com/njaremko/samael/issues/69).

dginev:
Fixes #165 .

Since you made the previous `wrapper.h` @wetneb , do you have any thoughts on extending the bindgen coverage to match all libxml2 headers available? As it turns out, other external projects had used some of them for convenience, despite not having a proper Rust API.


wetneb:
Sure! Indeed I only included the ones that were necessary for the Rust API. It would intuitively feel better not to expose any of the C bits and only let people rely on the Rust API, but I imagine there are plenty of parts of the API that aren't covered in Rust yet. On the long term, adopting the pattern of one public-facing crate with Rust API, and another `-sys` crate doing the binding, could be beneficial.

dginev:
Right, I am of two minds here myself, could see the point made either way.

Separating into a `-sys` crate has been requested actually, back in #76 . I was never fully convinced there is a long-term benefit in that refactor. Maybe that becomes more obvious if bindgen exposes a much larger surface than the one used by the Rust API... So I could continue this PR by separating the workspace into two crates and bumping to `v0.4.0`. Hm.

PR #173, opened 2025-07-08
Follow-up cleanup for canonicalization (c14n)

dginev:
Follow-up cleanup to #138


dginev:
Getting CI to pass on all operating systems - done.

PR #184, opened 2025-09-01
check if libxml2 includes are in a libxml2 subdirectory

zerocom38:
## Issue

With the latest update of libxml2 for Windows (vcpkg) to version 2.14.5, the include path has been moved to a `libxml2` subdirectory.

## Fix

If a `libxml2` subdirectory exists, switch to it. Otherwise, stay with the original include path.

## Note

The latest `cargo fmt` applied some format changes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "libxml"
-version = "0.3.7"
+version = "0.3.8"
 edition = "2024"
 authors = ["Andreas Franzén <andreas@devil.se>", "Deyan Ginev <deyan.ginev@gmail.com>","Jan Frederik Schaefer <j.schaefer@jacobs-university.de>"]
 description = "A Rust wrapper for libxml2 - the XML C parser and toolkit developed for the Gnome project"
--- a/build.rs
+++ b/build.rs
@@ -124,17 +124,19 @@ fn main() {
 mod vcpkg_dep {
   use crate::ProbedLib;
   pub fn vcpkg_find_libxml2() -> Option<ProbedLib> {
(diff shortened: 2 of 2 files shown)


dginev:
Thank you for the contribution!

Is there a way to induce the new Windows behavior in our Github CI? It would be nice to see the previous version fail and this one pass, to be sure the problem is addressed.

I can do that too since you've disclosed the version, but it will take me a while.

zerocom38:
I opened up an issue in vcpkg repo, if the subdirectory is expected or not: [https://github.com/microsoft/vcpkg/issues/47155](https://github.com/microsoft/vcpkg/issues/47155)

zerocom38:
> Thank you for the contribution!
> 
> Is there a way to induce the new Windows behavior in our Github CI? It would be nice to see the previous version fail and this one pass, to be sure the problem is addressed.
> 
> I can do that too since you've disclosed the version, but it will take me a while.

I'll check that.

zerocom38:
I did a lot of research. vcpkg recommends using cmake or pkg-config and not relying on the main include path. pkg-config is not available on Windows by default, so better not use it. The libxml2 includes seem to be in a libxml2 subdirectory on other platforms too (e.g Debian Linux). So I added the subdirectory to the include of vcpkg too. Even older versions of the vcpkg repo have the libxml2 subdirectory.

zerocom38:
@dginev Can we go ahead and merge this PR? There's still an ongoing discussion at vcpkg, but my change is just compatible with all versions. Right now, I cannot use the latest vcpkg version, because the libxml crate doesn't compile anymore.

dginev:
Looks good, thanks

PR #191, opened 2026-03-20
Add support for multi-threaded xpath evaluation

maartendeprez:
The `readonly` module provides support for multi-threaded read-only operations over a parsed document. However, the `findnodes` method takes a shared reference to a `Document`, which is not readonly and therefore not `Send` + `Sync`. This makes it impossible, currently, to evaluate xpaths over the document from multiple threads.

This PR changes that by introducing a read-only version of `Document`, `Context` and `Object`, as well as a `findnodes_readonly` method on `RoNode`.

### DIFF
--- a/src/readonly.rs
+++ b/src/readonly.rs
@@ -1,3 +1,9 @@
+mod context;
+mod document;
 mod tree;
+mod xpath;
 
-pub use self::tree::RoNode;
+pub use context::RoContext;
+pub use document::RoDocument;
+pub use tree::RoNode;
+pub use xpath::RoObject;
--- a/src/readonly/context.rs
+++ b/src/readonly/context.rs
@@ -18,8 +18,50 @@ impl RoContext {
     Ok(Self(context))
   }
 
+  /// evaluate an xpath
+  pub fn evaluate(&self, xpath: &str) -> Result<RoObject, ()> {
+    self.0.evaluate(xpath).map(RoObject)
+  }
+
+  ///evaluate an xpath on a context Node
+  pub fn node_evaluate(&self, xpath: &str, node: &RoNode) -> Result<RoObject, ()> {
+    self.0.node_evaluate_readonly(xpath, *node).map(RoObject)
+  }
+
   /// evaluate an xpath on a context RoNode
   pub fn node_evaluate_readonly(&self, xpath: &str, node: RoNode) -> Result<RoObject, ()> {
     self.0.node_evaluate_readonly(xpath, node).map(RoObject)
+  }
+
+  /// find nodes via xpath, at a specified node or the document root
+  pub fn findnodes(&self, xpath: &str, node_opt: Option<&RoNode>) -> Result<Vec<RoNode>, ()> {
+    // Note: we cannot implemented this as `self.0.findnodes(...)` because that
+    // method takes `&mut self`.
+    let evaluated = if let Some(node) = node_opt {
+      self.node_evaluate(xpath, node)?
+    } else {
+      self.evaluate(xpath)?
+    };
+    Ok(evaluated.get_nodes_as_vec())
+  }
+
+  /// find literal values via xpath, at a specified node or the document root
+  pub fn findvalues(&self, xpath: &str, node_opt: Option<&RoNode>) -> Result<Vec<String>, ()> {
+    let evaluated = if let Some(node) = node_opt {
+      self.node_evaluate(xpath, node)?
+    } else {
+      self.evaluate(xpath)?
+    };
+    Ok(evaluated.get_nodes_as_str())
+  }
+
+  /// find a literal value via xpath, at a specified node or the document root
+  pub fn findvalue(&self, xpath: &str, node_opt: Option<&RoNode>) -> Result<String, ()> {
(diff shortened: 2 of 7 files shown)


dginev:
Thank you, this appears to be a very useful extension.

Could I also request that you mirror our one findnodes test here:
https://github.com/KWARC/rust-libxml/blob/4224fd91cc0d1e8f96651606dfeb2436bc589bb5/tests/xpath_tests.rs#L122-L144

With a similar entry inside `readonly_tests.rs` ?

maartendeprez:
Copied and adapted the tests and added the required read-only methods to implement this. I left two tests commented out in the read-only version:

- `cleanup_safely_unlinked_xpath_nodes`
- `xpath_with_namespaces`

Both require the `register_namespace` method, which seems to require exclusive access to the `RoContext` object. This could be ensured by taking `&mut self`, were it not that the context is cloneable without actually duplicating the underlying libxml object. The first one would also require `RoDocument::as_node`, which requires exclusive access to the document.

maartendeprez:
Hm, upon further consideration, my design is not correct at all, since the `Ro*` types wrap the read-write variants containing `Rc` to the libxml pointer, which lacks the necessary atomicity in the reference counter to share this between threads. I'll need to rethink this.

dginev:
> I'll need to rethink this.

Is there anything I can help with? The main background I can offer is that multi-threaded workflows on the `Ro*` nodes are completely OK as long as they only ever need to do read-only operations.

Mixing the read-only API of the create with the writable API is likely something that needs some significant redesign work, as it wasn't an original desire. My thinking at the time was, if you need even one writable operation, do not use the `Ro*` types.

maartendeprez:
My thinking at the moment is that for the read-only API, I'll need to do away with all the reference counting stuff (i.e. the `Rc`s) in the `Ro*` types and use the auto-generated pointer types directly, as you did for `RoNode`. Unfortunately that causes quite a bit of code duplication.

Excluding mixed use of the writable and read-only APIs at the type level, is what i was alluding to in [issue 192](https://github.com/KWARC/rust-libxml/issues/192). That would need to be done in a separate PR. In rust when we're "not supposed to" do something that would cause undefined behavior, the API is usually designed in such a way that the compiler will not allow us to do so without writing "unsafe" code, ensuring that ["Safe Rust can’t cause Undefined Behavior"](https://doc.rust-lang.org/nomicon/safe-unsafe-meaning.html).

On the other hand, if we want to allow mixing read-only and read-write APIs, it seems possible to leverage the rust type-system to enforce the contract on top of the `Ro*`-style API. Methods that require exclusive access would take `&mut self`, forcing the user to take care of exclusion rules, e.g. by wrapping a shared `RoDocument` in a `Mutex` or `RwLock` if it needs to be accessed from multiple threads. However I am not intimately familiar with `libxml2`; you might immediately see issues I am not aware of.

If that kind of API is feasible, it would provide a completely different (arguably lower-level) API. If we go through with this, I think we should separate the bindings in a `libxml2-sys` crate and provide two separate crates for the two styles of "safe" API wrappers.

maartendeprez:
I have been [experimenting](https://github.com/maartendeprez/rust-libxml/tree/lower-level-safe-api) with a statically checked mutable / read-only API and i believe i am starting to understand the issues for the mutable API. Since every node is  connected to its parent, children and siblings, the whole tree is reachable from every node. This means we need to be careful what combination of methods are allowed for gaining exclusive access to a node. For example, i believe we could either provide:

- `get_first_child_mut`, `get_next_sibling_mut`, `get_parent_mut`, ...
- `iter_children_mut` (see [this comment](https://github.com/maartendeprez/rust-libxml/blob/ef5cb6336f900c9dabe85ed0e02889158035e89b/libxml-core/src/tree/node.rs#L172)) 

Providing both would allow obtaining multiple exclusive references to the same node, since the mutable iterator provides split borrows. Some ideas on how to improve this:

- Return wrapped references from mutable iterators; the wrapper would disallow calling methods that provide access to neighboring nodes
- Provide a [`CursorMut`-like API](https://doc.rust-lang.org/std/collections/linked_list/struct.CursorMut.html), where permission to walk over the tree is borrowed from the collection (document or owned node)

I believe the second option would also solve issues with implementing `NodeRef::unlink(&mut Self) -> Node`.


All-in-all, i believe the statically checked API would end up considerably more restrictive than the current `Rc`-based one in terms of concurrent mutable access. It would, of course, improve the situation for multi-threaded usage. The (mostly) read-only part seems uncomplicated though.

Let me know your thoughts. Would you be on board with adding the new API as an experimental crate in this repo? Alternatively, i could make a separate PR for separating out the `libxml-sys` crate.

dginev:
> Let me know your thoughts. 

It sounds ambitious! But honestly, it also sounds too close to not needing a wrapper at all and implementing an outright Rust library for XML manipulation - as many other projects do.

The allure of wrapping over libxml2 is that we get something mature and high-performance, while we accept it is not native Rust and will have (well-motivated) limitations for safety and ownership.

So I am somewhat wondering if this is too ambitious and risks the usefulness of the wrapper for its more pragmatic aim. Why not use a pure Rust XML library once your needs get to that level?

PR #193, opened 2026-03-22
Using libxml2 source code

MaxKingPor:


dginev:
@MaxKingPor would you like to add a detailed description here? Are you trying to bundle the original libxml2 source code into the Rust wrapper crate?

As you can see from the project README that is not something we want to do in this crate - which aims to be a wrapper on top of a preinstalled libxml2 distribution.

The one interesting exception would be if we can use c2rust (or a similar project) to stabilize a Rust translation of libxml2. But even that may fit better as an experimental counterpart to this crate.

For now I will close this PR, and you are welcome to fork the project and experiment as you wish. But please double-check the licensing of libxml2 allows for the kinds of redistribution you are interested in.

PR #194, opened 2026-04-22
Node::new_comment and init_parser guard

dginev:
Two small high-level wrappers:

* `Node::new_comment(content, doc)` — mirrors `new_text`, wraps `xmlNewDocComment`. Returns `Err(())` if the content contains a NUL or libxml2 returns NULL. Returned node is unlinked; callers attach via `add_child` / `add_prev_sibling` / `add_next_sibling`.

* `libxml::init_parser()` at the top-level — Once-guarded wrapper around `xmlInitParser`. Safe to call from multiple threads; the underlying C init runs exactly once. Useful for applications that don't go through `Parser::default()` (which already initialises lazily) but still need libxml2 globals available — e.g. when constructing `Document` or `Node` handles directly at startup.

We'll also publish v0.3.9 with this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,23 @@
 # Change Log
 
-## [0.3.9] (in development)
+## [0.3.10] (in development)
+
+## [0.3.9] (2026-04-22)
+
+### Added
+
+* `Node::new_comment` constructor, wrapping `xmlNewDocComment`. Returns an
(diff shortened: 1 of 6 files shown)


dginev:
These were needed in the latexml Rust rewrite we are working on, and were prepared in part with the help of Opus 4.7. Times are changing.

dginev:
I'll merge this for now and we can adjust later if needed.

